PR #170, opened 2026-03-03
Load source metadata from nexus

jl-wynen:
So far, we have been hard-coding the source in workflows that use it. But the coda nexus files now contain enough information to load this instead.

This will help with https://github.com/scipp/essdiffraction/issues/100

### DIFF
--- a/packages/essreduce/src/ess/reduce/nexus/workflow.py
+++ b/packages/essreduce/src/ess/reduce/nexus/workflow.py
@@ -15,6 +15,7 @@ import scippnexus as snx
 from scipp.constants import g
 from scipp.core import label_based_index_to_positional_index
 from scippneutron.chopper import extract_chopper_from_nexus
+from scippneutron.metadata import RadiationProbe, SourceType
 
 from . import _nexus_loader as nexus
 from .types import (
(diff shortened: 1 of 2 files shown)


packages/essreduce/src/ess/reduce/nexus/workflow.py
```diff
@@ -629,6 +631,31 @@ def load_measurement_metadata_from_nexus(
...
 
 
+def load_source_metadata_from_nexus(
+    source_component: NeXusComponent[snx.NXsource, RunType], beamline: Beamline[RunType]
+) -> Source[RunType]:
+    """Load source metadata from a NeXus file.
+
+    Supplements the result with data from ``beamline`` if necessary.
+    """
+    from scippneutron.metadata import RadiationProbe, SourceType
```
MridulS: Any particualr reason for this inline import?

jl-wynen: I thought I'd keep the scippneutron types hidden (as opposed to `Source[T]` which is defined in ESSreduce. But that should not be an issue

MridulS (approved):

PR #172, opened 2026-03-06
Add report job status step to CI workflow

YooSunYoung:
With the matrix strategy, we couldn't use them for branch protection rules.
Here I added the final report job that fails when `formatting`, `test` or `docs` job fails, and not skipped.

We can use this one for the branch protection rule maybe...?

I tested the failure report in #173


YooSunYoung:
Is it not so useful after all cause we will see failing jobs anyway...?

nvaytet:
I think this is useful :+1:

PR #176, opened 2026-03-06
Copy over LICENSE file manually instead of a relative link

MridulS:


packages/essimaging/pyproject.toml
```diff
@@ -10,7 +10,6 @@ name = "essimaging"
 description = "Imaging data reduction for the European Spallation Source"
 authors = [{ name = "Scipp contributors" }]
 license = "BSD-3-Clause"
-license-files = ["../../LICENSE"]
 readme = "README.md"
```
MridulS, line 13: The `"../../LICENSE" relative path confuses setuptools, so we can just copy over the LICENSE to bundle it with the wheels.

PR #177, opened 2026-03-06
Use pixi install frozen instead of locked in CI

MridulS:


nvaytet:
Can you explain quickly the gist of the changes?

MridulS:
We have editable installs which means it always tries to relock the deps even though there is already a pixi.lock in the root. The frozen part makes sure that the CI doesn't try to relock the deps.

nvaytet (approved):

PR #184, opened 2026-03-09
Fix check for empty detectors

jl-wynen:
Looks like this was not transferred from the old repo. Original: https://github.com/scipp/essreduce/pull/320

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -10,28 +10,105 @@ environments:
     packages:
       linux-64:
       - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-20_gnu.conda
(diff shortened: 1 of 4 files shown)


pixi.lock
nvaytet, line 1: I don't think the pixi lock file should be included.

jl-wynen: Why not? I changed the package dependencies.

nvaytet: ah yes indeed...
Problem with the lock file is that it's so large that I don't even look at the changes.
Before, I still looked at the requirements files and sometimes spotted things that had gone wrong.
Here, I feel we are blindly trusting pixi to do the right thing.

nvaytet (approved):

PR #185, opened 2026-03-09
Simplify tag_regex a bit

MridulS:
I think the more strict regex seems to create issues on certain systems. This should also make it easier to debug.

### DIFF
--- a/packages/essimaging/pyproject.toml
+++ b/packages/essimaging/pyproject.toml
@@ -72,7 +72,7 @@ docs = [
 
 [tool.setuptools_scm]
 root = "../.."
-tag_regex = "^essimaging/(?P<version>[vV]?\\d+(?:\\.\\d+)*(?:[._-]?\\w+)*)$"
+tag_regex = "^essimaging/(?P<version>.*)$"
 git_describe_command = ["git", "describe", "--dirty", "--tags", "--long", "--match", "essimaging/*[0-9]*"]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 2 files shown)


packages/essimaging/pyproject.toml
```diff
@@ -72,7 +72,7 @@ docs = [
 
 [tool.setuptools_scm]
 root = "../.."
-tag_regex = "^essimaging/(?P<version>[vV]?\\d+(?:\\.\\d+)*(?:[._-]?\\w+)*)$"
+tag_regex = "^essimaging/(?P<version>.*)$"
```
jl-wynen: Not sure it should be this loose. Can you give an example why we can't have stricter requirements on the tag? At least something like `^essimaging/(?P<version>[vV]?\d+.*)$ to ensure the tag starts with a number (or `v<number>`).

MridulS: I'm fine with that too. I just wanted to remove the escapes as much as possible. I'm not sure if these things affects different systems differently?

jl-wynen: Which escapes are you talking about?

MridulS: The double `\\`

YooSunYoung: `^essimaging/(?P<version>[vV]?\\d+.*)$`

Like this right?

We decided not to use `v` so shouldn't it be like this?

`^essimaging/(?P<version>\\d+.*)$`

jl-wynen: > The double \\

This should be independent of the OS. It's Python-specific.

> `^essimaging/(?P<version>[vV]?\\d+.*)$`
>
> Like this right?

Yes.

> We decided not to use v so shouldn't it be like this?

No harm allowing a v IMHO. But if we want to be strict, then yes, we can leave it out.

jl-wynen (approved):

PR #243, opened 2026-03-12
Unpinned test in nightly.

YooSunYoung:
I tried running the unpinned test manually:

https://github.com/scipp/ess/actions/runs/23005201613/job/66799997990

`essnmx` fails as expected but `essreduce` also fails, which was not expected...
Do you know why it fails...? @nvaytet


nvaytet:
I have no idea :disappointed:

YooSunYoung:
Maybe unpinned tests are not necessary anymore 
since our dev environment always install dev version of the dependency(essreduce)
and tests get updated earlier and updated tests are not necessarily backward compatible...?

nvaytet:
The fact that sub-package tests are run when `essreduce` is updated, plus the fact that we will find a way to make 'batch releases' where we release all packages in one go, means that this doesn't seem relevant anymore.

PR #246, opened 2026-03-16
Add customisation hook for loading time-dependent transformations from NeXus

jl-wynen:
This PR adds a hook to customise how time-dependent transformations are loaded from NeXus and merged with the data. This PR should not break any existing code. It only adds new options to the workflow.

See https://github.com/scipp/essspectroscopy/pull/101 for a use case.

### DIFF
--- a/packages/essreduce/src/ess/reduce/nexus/workflow.py
+++ b/packages/essreduce/src/ess/reduce/nexus/workflow.py
@@ -8,6 +8,7 @@ from collections.abc import Iterable
 from copy import deepcopy
 from typing import Any, TypeVar
 
+import numpy as np
 import sciline
 import sciline.typing
 import scipp as sc
@@ -275,26 +276,51 @@ def load_nexus_data(
 
 
 def get_transformation_chain(
-    detector: NeXusComponent[Component, RunType],
+    component: NeXusComponent[Component, RunType],
 ) -> NeXusTransformationChain[Component, RunType]:
     """
-    Extract the transformation chain from a NeXus detector group.
+    Extract the transformation chain from a NeXus component group.
 
     Parameters
     ----------
-    detector:
-        NeXus detector group.
+    component:
+        NeXus component group.
     """
-    chain = detector['depends_on']
+    chain = component['depends_on']
     return NeXusTransformationChain[Component, RunType](chain)
 
 
-def _time_filter(transform: sc.DataArray) -> sc.Variable:
+def _collapse_runs(transform: sc.DataArray, dim: str) -> sc.DataArray:
+    """Collapse runs of equal values into a single value."""
+    # Find indices where the data changes
+    different_from_previous = np.hstack(
+        [True, ~np.isclose(transform.values[:-1], transform.values[1:])]
+    )
+    change_indices = np.flatnonzero(different_from_previous)
+    if change_indices.shape == transform.shape:
+        return transform  # Return early to avoid expensive indexing
+    # Get unique values
+    unique_values = transform[change_indices]
+
+    # Make bin-edges and extend range to include the whole measurement
+    last = unique_values.coords[dim][-1]
+    unique_values.coords[dim] = sc.concat(
+        [
+            sc.epoch(unit=last.unit),
+            unique_values.coords[dim][1:],
+            # Surely, no experiment will last more than 10 years...
+            last + sc.scalar(10, unit='Y').to(unit=last.unit),
+        ],
+        dim=dim,
+    )
+
+    return unique_values
+
+
+def _time_filter(transform: sc.DataArray) -> sc.Variable | sc.DataArray:
     if transform.ndim == 0 or transform.sizes == {'time': 1}:
         return transform.data.squeeze()
-    raise ValueError(
-        f"Transform is time-dependent: {transform}, but no filter is provided."
-    )
+    return _collapse_runs(transform, dim='time')
 
 
 def to_transformation(
@@ -369,6 +395,10 @@ def get_calibrated_detector(
     The data array is reshaped to the logical detector shape, by folding the data
     array along the detector_number dimension.
 
+    The output contains pixel positions computed from ``transform`` and ``offset``.
+    If ``transform`` is time-dependent, the output contains a 'time' dimension
+    and coordinate corresponding to the time coordinate of ``transform``.
+
     Parameters
     ----------
     detector:
@@ -401,9 +431,17 @@ def get_calibrated_detector(
     else:
         transform_value = transform.value
     position = transform_value * offsets
-    return EmptyDetector[RunType](
-        da.assign_coords(position=position + offset.to(unit=position.unit))
-    )
+
+    position = position + offset.to(unit=position.unit)
+    if isinstance(position, sc.DataArray):  # time-dependent transform
+        # Store position and time as separate coords because we can't store data arrays.
+        return EmptyDetector[RunType](
+            da.broadcast(
+                dims=['time', *da.dims], shape=[position.sizes['time'], *da.shape]
+            ).assign_coords(position=position.data, time=position.coords['time'])
+        )
+
+    return EmptyDetector[RunType](da.assign_coords(position=position))
 
 
 def assemble_detector_data(
@@ -422,13 +460,30 @@ def assemble_detector_data(
     neutron_data:
         Neutron data array (events or histogram).
     """
-    if neutron_data.bins is not None:
+    detector_coords = dict(detector.coords)
+    position = detector_coords.get('position')
+    if neutron_data.is_binned:
         neutron_data = nexus.group_event_data(
             event_data=neutron_data, detector_number=detector.coords['detector_number']
         )
+        if position is not None and 'time' in position.dims:
+            del detector_coords['position']
+            time = detector_coords.pop('time')
+            pos_lookup = sc.lookup(
+                sc.DataArray(position, coords={'time': time}), dim='time'
+            )
+            neutron_data.bins.coords['position'] = pos_lookup[
+                neutron_data.bins.coords['event_time_zero']
+            ]
+    else:
+        if position is not None and 'time' in position.dims:
+            raise NotImplementedError(
+                "Time-dependent positions are not yet supported for histogram data."
+            )
+
     return RawDetector[RunType](
         _add_variances(neutron_data)
-        .assign_coords(detector.coords)
+        .assign_coords(detector_coords)
         .assign_masks(detector.masks)
     )
 
@@ -658,7 +713,6 @@ def load_source_metadata_from_nexus(
 definitions = snx.base_definitions()
 definitions["NXdetector"] = _StrippedDetector
 definitions["NXmonitor"] = _StrippedMonitor
-
 
 _common_providers = (
     gravity_vector_neg_y,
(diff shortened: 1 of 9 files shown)


jl-wynen:
This causes an issue with the `Position` domain type. It used to always be a (scalar) Variable. Now, when a component moves, its `Position` is a 1d data array. That means that the type should be
```python
class Position(sciline.Scope[Component, RunType, sc.Variable | sc.DataArray], sc.Variable | sc.DataArray):
```
But that is not possible because we can't inherit from unions. So I'm thinking we could always make positions data arrays. But that impacts a lot of downstream code that currently expects variables. But that code is anyway broken for dime dependent positions and might need an update.

Alternatively, we could use a separate path through the workflow and add
```python
class MovingPosition(sciline.Scope[Component, RunType, sc.DataArray], sc.DataArray):
```
and providers then have to request that type when they can deal with time dependence. The providers constructing `Position` would then raise if they get time-dependent inputs.

Thoughts? (@SimonHeybrock, @nvaytet)

packages/essreduce/src/ess/reduce/nexus/workflow.py
```diff
@@ -347,9 +366,22 @@ def compute_position(
...
+            "transformation. Either provide a time interval parameter or "
+            "time filter."
+        )
     return Position[Component, RunType](transformation.value * origin)
 
 
+def compute_dynamic_position(
+    transformation: NeXusTransformation[Component, RunType],
+) -> DynamicPosition[Component, RunType]:
+    """Compute the position of a component from a transformation matrix."""
```
nvaytet, line 381: Nitpick: Should there also be an error message if we attempt here to compute a dynamic position from a scalar variable?

packages/essreduce/src/ess/reduce/nexus/workflow.py
```diff
@@ -712,11 +754,14 @@ def LoadMonitorWorkflow(
     """Generic workflow for loading monitor data from a NeXus file."""
     wf = sciline.Pipeline(
         (*_common_providers, *_monitor_providers),
+        params={
+            PreopenNeXusFile: PreopenNeXusFile(False),
+            TransformationTimeFilter: reject_time_dependent_transform,
+        },
```
nvaytet, line 760: Can you explain/sketch out what the e.g. Loki workflow would look like if we need to load both moving and stationary monitors?
I didn't see where the `DynamicPosition` is being used.

packages/essreduce/src/ess/reduce/unwrap/to_wavelength.py
```diff
@@ -398,8 +398,8 @@ def detector_ltotal_from_straight_line_approximation(
     # TODO: scatter=True should not be hard-coded here
     graph = {
         **scn.conversion.graph.beamline.beamline(scatter=True),
-        'source_position': lambda: source_position,
-        'sample_position': lambda: sample_position,
+        'source_position': lambda: source_position.position,
+        'sample_position': lambda: sample_position.position,
```
nvaytet: I'm wondering if this can potentially break a lot of existing code?

jl-wynen: Yes. That should not have been there anymore. I guess I forgot to push something. See the update.

For context: This PR should not break any existing code. It loads 'static' positions as before using the `Position` type. It adds the `DynamicPosition` type that allows for potentially time-dependent data. But you need to explicitly request that in the workflow. (Used for BIFROST analysers)

packages/essreduce/src/ess/reduce/nexus/workflow.py
jl-wynen: > I didn't see where the DynamicPosition is being used.

It's unused here. But look at this PR: https://github.com/scipp/essspectroscopy/

As for Loki: You don't need to do anything for stationary monitors. Let's say, `Monitor2` moves. Then you need to set
```python
def monitor_time_filter(transform: sc.DataArray) -> sc.DataArray: ...
def get_calibrated_dmoving_monitor(
    monitor: NeXusComponent[Monitor2, RunType],
    *,
    transform: NeXusTransformation[Monitor2, RunType],
) -> EmptyMonitor2[Monitor2, RunType]: ...

wf[TransformationTimeFilter[Monitor2, RunType]] = monitor_time_filter
wf.insert(get_calibrated_dmoving_monitor)
```
This overrides the default for `Monitor2` and allows you to specify how to handle the time-dependence.

packages/essreduce/src/ess/reduce/nexus/workflow.py
jl-wynen: I would say no. `DynamicPosition` can still be stationary. E.g., the BIFROST workflow should transparently handle moving and stationary detectors without changes to the graph.

packages/essreduce/src/ess/reduce/nexus/workflow.py
```diff
@@ -275,33 +277,47 @@ def load_nexus_data(
...
+    chain = component['depends_on']
     return NeXusTransformationChain[Component, RunType](chain)
 
 
-def _time_filter(transform: sc.DataArray) -> sc.Variable:
-    if transform.ndim == 0 or transform.sizes == {'time': 1}:
-        return transform.data.squeeze()
+def reject_time_dependent_transform(
+    transform: sc.DataArray,
+) -> Never:
```
SimonHeybrock, line 296: Didn't know `Never`, but sources say `NoReturn` is the same but might be preferable in return types?

packages/essreduce/src/ess/reduce/nexus/workflow.py
```diff
@@ -336,9 +355,9 @@ def to_transformation(
             idx = label_based_index_to_positional_index(
                 sizes=t.sizes, coord=time, index=interval.value
             )
-            t.value = _time_filter(t.value[idx])
+            t.value = _time_filter(t.value[idx], time_filter)
```
SimonHeybrock: Confusing near-duplicate variable naming. Should `_time_filter` by `_apply_time_filter` or something like that?

packages/essreduce/src/ess/reduce/nexus/workflow.py
jl-wynen: They are the same https://docs.python.org/3/library/typing.html:

> Never and NoReturn have the same meaning in the type system and static type checkers treat both equivalently.

packages/essreduce/src/ess/reduce/nexus/workflow.py
```diff
@@ -347,9 +366,22 @@ def compute_position(
...
+        raise ValueError(
+            "Attempted to compute a static position from a time-dependent "
+            "transformation. Either provide a time interval parameter or "
+            "time filter."
+        )
     return Position[Component, RunType](transformation.value * origin)
 
 
+def compute_dynamic_position(
+    transformation: NeXusTransformation[Component, RunType],
```
SimonHeybrock, line 379: The `to_transformation` docstring returning this still claims it is time-independent?

packages/essreduce/src/ess/reduce/nexus/workflow.py
```diff
@@ -670,11 +712,11 @@ def load_source_metadata_from_nexus(
     get_transformation_chain,
     to_transformation,
     compute_position,
+    compute_dynamic_position,
```
SimonHeybrock, line 715: How does this work? The providers do not look specialized (both use `Component` typevar), what does it mean to have both here?

packages/essreduce/src/ess/reduce/nexus/workflow.py
nvaytet: Why do we need to do 2 things to the workflow? (1. set `TransformationTimeFilter`, 2. insert a new provider)
Can the filtering just be done inside the provider?

Or maybe it's because inserting the provider might be something added in e.g. `essspectroscopy` inside the module, but the time filter may be tweaked by the user in the notebook, hence the separation?

packages/essreduce/src/ess/reduce/nexus/workflow.py
jl-wynen: This allows you to request either `Position` or `DynamicPosition` and sciline picks the matching provider.

packages/essreduce/src/ess/reduce/nexus/workflow.py
jl-wynen: Which line are you referring to? `to_transformation` can indeed return a time-dependent transform. But that depends on `TransformationTimeFilter`.

packages/essreduce/src/ess/reduce/nexus/workflow.py
SimonHeybrock: @jl-wynen regarding https://github.com/scipp/ess/pull/246#discussion_r3098556384

packages/essreduce/tests/nexus/workflow_test.py
```diff
@@ -172,16 +172,6 @@ def test_to_transform_with_label_based_time_interval_single_point(
     assert sc.identical(transform * origin, sc.vector([1.0, 3.0, 0.0], unit='m'))
 
 
-def test_to_transform_raises_if_interval_does_not_yield_unique_value(
```
nvaytet, line 175: Do I remember correctly that there used to be some tests that you added in this PR, but now it seems they are gone?
Or am I somehow reading the wrong diff?

jl-wynen: They are gone. I could add them back but that requires implementing the custom function I outlined in the Loki comment which makes the tests somewhat complicated. Should I do that? Or is it enough to have the bifrost tests which will soon be in the monorepo?

packages/essreduce/src/ess/reduce/nexus/workflow.py
jl-wynen: Ah, right. Fixed it

packages/essreduce/src/ess/reduce/nexus/workflow.py
nvaytet: > But look at this PR: https://github.com/scipp/essspectroscopy/

Can you provide a more specific link? :wink:

packages/essreduce/tests/nexus/workflow_test.py
nvaytet: I guess if one day someone decides to remove the tests in Bifrost, we may forget that `reduce` was relying on those, and we would then no longer be testing this anywhere?

I guess it depends how complicated the tests get...

SimonHeybrock (approved):

packages/essreduce/src/ess/reduce/nexus/workflow.py
jl-wynen: This is the filter: https://github.com/scipp/essspectroscopy/blob/6bee451edceef4c9b0097180d064a79f2dec53f9/src/ess/bifrost/io/nexus.py#L178
This is how the data is combined: https://github.com/scipp/essspectroscopy/blob/6bee451edceef4c9b0097180d064a79f2dec53f9/src/ess/bifrost/detector.py#L83

SimonHeybrock:
@jl-wynen Before merging, please also update the PR description!

packages/essreduce/tests/nexus/workflow_test.py
jl-wynen: True. I restored a test I accidentally removed and added one more. It only tests `to_transformation`. Do we need a full workflow test? That would mean making the input data.

nvaytet (approved):

PR #248, opened 2026-03-17
[Essimaging]: Do not allow nbsphinx errors in docs builds

nvaytet:
For some reason, we had `nbsphinx_allow_errors = True` enabled in the docs sphinx config.

**Note more changes here:** we set the `LookupTableRelativeErrorThreshold` to a default value of `Inf` in Odin and TBL workflows, to avoid breaking all existing notebooks.

### DIFF
--- a/packages/essimaging/tests/odin/data_reduction_test.py
+++ b/packages/essimaging/tests/odin/data_reduction_test.py
@@ -3,7 +3,6 @@
 
 import pytest
 import sciline as sl
-from ess.reduce.time_of_flight import LookupTableRelativeErrorThreshold
 
 import ess.odin.data  # noqa: F401
 from ess import odin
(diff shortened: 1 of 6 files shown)


packages/essimaging/tests/odin/data_reduction_test.py
```diff
@@ -30,9 +29,6 @@ def workflow() -> sl.Pipeline:
     wf[Filename[OpenBeamRun]] = odin.data.iron_simulation_ob_small()
     wf[NeXusDetectorName] = "event_mode_detectors/timepix3"
     wf[TimeOfFlightLookupTableFilename] = odin.data.odin_tof_lookup_table()
-    wf[LookupTableRelativeErrorThreshold] = {
```
nvaytet, line 33: These are now set as default in the workflow, so it is no longer needed here.

YooSunYoung (approved):

PR #249, opened 2026-03-17
Fix tof resampling for single slice

jl-wynen:
This makes the cases with a single slice and multiple slices consistent.


nvaytet (approved):

nvaytet:
@SimonHeybrock FYI we found a bug in the rebinning of strictly increasing sections

Also, the tests were not run because the file name ended with `tests` instead of `test` :roll_eyes:

PR #250, opened 2026-03-17
Improve docs build ergonomics

jl-wynen:
This PR makes a couple of changes:
- It adds missing paths to gitignore
- It moves the doctrees into the packages so that they don't collide with each other in `/tmp` (Plus, `/tmp` is a Unix path)
- It adds a separate docs task for each package. Now we can run the docs using
```shell
pixi run docs-essimaging
```

Unfortunately, I could not find a way to reuse the sphinx command and also automatically activate the right environment.

## How this works
Take for example ESSreduce: We have a `docs-essreduce` _feature_. This feature contains a _task_ also called `docs-essreduce`. This connects the task to that feature. Since we only have one _environment_ (also called `ess-reduce`) that contains the `docs-essreduce` _feature_, pixi knows to use that environment for this task. (And all tasks in this feature.)
This can be confusing, especially since there are 3 different things with the same name :/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
           frozen: true
           environments: docs-${{ matrix.package }}
       - name: Build docs
-        run: pixi run -e docs-${{ matrix.package }} docs ${{ matrix.package }}
+        run: pixi run docs-${{ matrix.package }}
       - uses: actions/upload-artifact@v4
         id: artifact-upload-step
         with:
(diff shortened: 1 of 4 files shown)


jl-wynen:
Correction: I had to add `default-environment` to the `docs-essreduce` _task_ because the `docs-essreduce` _feature_ appears in all `docs-*` _environments_ and pixi was unable to figure out which env to use on its own.

nvaytet (approved):

PR #253, opened 2026-03-18
Move helper functions to conftest.

YooSunYoung:
I moved the non-test-python-file to conftest.
It's better this way I think?


packages/essnmx/tests/conftest.py
jl-wynen, line 1: Can you use more explicit fixture names? `no_detectors` doesn't say anything about what type of data it contains. And you can use a `conftest.py` in the `mcstas` subfolder to give the fixtures a more narrow scope.

jl-wynen:
Maybe you should base this branch on main. That way, CI doesn't fail because of the file name in essreduce.

packages/essnmx/tests/conftest.py
YooSunYoung: Good point...

YooSunYoung:
I'll close this in favor of #255

PR #256, opened 2026-03-18
Move ruff config to root, remove duplicate config

MridulS:


ruff.toml
```diff
@@ -0,0 +1,38 @@
+line-length = 88
```
YooSunYoung, line 1: ~Can we increase this limit to 128 instead...? It makes the notebooks super ugly....~ Actually it was in the ignore list

ruff.toml
```diff
@@ -0,0 +1,38 @@
...
+pydocstyle.convention = "numpy"
+
+[lint.per-file-ignores]
+# those files have an increased risk of relying on import order
+"packages/*/tests/*" = [
+    "S101",  # asserts are fine in tests
+    "B018",  # 'useless expressions' are ok because some tests just check for exceptions
+]
+"*.ipynb" = [
+    "E501",  # longer lines are sometimes more readable
```
YooSunYoung, line 26: I think this ignore rule was ignored??

YooSunYoung (dismissed):
Notebooks shouldn't have the fixes related to E501 but other than that, it's fine...?

packages/essimaging/docs/odin/odin-data-reduction.ipynb
```diff
@@ -201,8 +201,14 @@
    "outputs": [],
    "source": [
     "wf[MaskingRules] = {\n",
-    "    'x_pixel_offset': lambda x: (x < sc.scalar(-5.8e-3, unit='m').to(unit=x.unit)) | (x > sc.scalar(5.8e-3, unit='m').to(unit=x.unit)),\n",
-    "    'y_pixel_offset': lambda y: (y < sc.scalar(-5.8e-3, unit='m').to(unit=y.unit)) | (y > sc.scalar(5.8e-3, unit='m').to(unit=y.unit))\n",
+    "    'x_pixel_offset': lambda x: (\n",
```
nvaytet: I think we had decided (probably a long time ago) to not enforce formatting on notebooks.
If people want to format them with ruff or black, they are free to do so.
But if they want to have long lines like these, it should be allowed.

packages/essimaging/docs/odin/odin-data-reduction.ipynb
```diff
@@ -238,10 +244,9 @@
    "source": [
     "wbins = sc.linspace('wavelength', 1.1, 9.4, 301, unit='angstrom')\n",
     "\n",
-    "normalized = (\n",
-    "        wf.compute(CorrectedDetector[SampleRun]).bins.concat().hist(wavelength=wbins) /\n",
-    "        wf.compute(CorrectedDetector[OpenBeamRun]).bins.concat().hist(wavelength=wbins)\n",
-    ")\n",
+    "normalized = wf.compute(CorrectedDetector[SampleRun]).bins.concat().hist(\n",
```
nvaytet: Same as above, I think it was more readable before.

packages/essimaging/docs/odin/odin-make-tof-lookup-table.ipynb
```diff
@@ -42,7 +42,9 @@
     "wf = time_of_flight.TofLookupTableWorkflow()\n",
     "wf[time_of_flight.DiskChoppers[AnyRun]] = disk_choppers\n",
     "wf[time_of_flight.SourcePosition] = source_position\n",
-    "wf[time_of_flight.NumberOfSimulatedNeutrons] = 200_000  # Increase this number for more reliable results\n",
+    "wf[time_of_flight.NumberOfSimulatedNeutrons] = (\n",
+    "    200_000  # Increase this number for more reliable results\n",
```
nvaytet: Here also.

ruff.toml
MridulS: We never really ran ruff format on our notebooks https://github.com/scipp/copier_template/blob/288894294bf0085bedcb275e5d5788aadfe3a619/template/.pre-commit-config.yaml#L31 and with this I didn't turn it off.

Added config to exclude .ipynb files from ruff format below.

packages/essnmx/pyproject.toml
```diff
@@ -143,11 +106,3 @@ enable_error_code = [
     "truthy-bool",
 ]
 warn_unreachable = true
-
-[tool.codespell]
```
nvaytet, line 147: Do we need to add this back into the global config?

MridulS: It's already there    https://github.com/scipp/ess/blob/f6b15b209101389157af810e5934e05e1d783e65/.pre-commit-config.yaml#L36

nvaytet (approved):

PR #257, opened 2026-03-18
Add security policy

jl-wynen:
While it is unlikely that we have security vulnerabilities in our code, we should still add this file as clarification. 
I copied this from Scitacean and it's based on a template I found a while ago. I'm open to modifications.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,19 @@
+# Security Policy
+
+## Supported Versions
+
+Only the latest version is supported.
+
+## Reporting a Vulnerability
+
+If you believe you have found a security vulnerability in an ess package, please
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,19 @@
+# Security Policy
+
+## Supported Versions
+
+Only the latest version is supported.
+
+## Reporting a Vulnerability
+
+If you believe you have found a security vulnerability in an ess package, please
```
MridulS, line 9:
```suggestion
If you believe you have found a security vulnerability in an ESS package, please
```

jl-wynen: I wanted to distinguish it from the facility. This policy only applies to this repo, not all ESS packages.

MridulS (approved):

PR #261, opened 2026-03-19
Add dev feature with jupyter

jl-wynen:
THis makes it possible to launch jupyter directly in a pixi env. 
For me personally, this allows running juptyer in pycharm. BUt it should also help with vs code.


YooSunYoung (approved):

YooSunYoung:
> BUt it should also help with vs code.

Yes it does...!

PR #262, opened 2026-03-24
Fix docs release workflow

nvaytet:
The change in the [docs build](https://github.com/scipp/ess/pull/250/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR98)
Screenshot_20260324_135717

was not propagated everywhere.

Here we fix it, and also change the `docs-deploy.yml` workflow to just build for a selected package.


.github/workflows/docs-deploy.yml
```diff
@@ -2,56 +2,55 @@ name: Deploy docs
 
 on:
   workflow_dispatch:
+    inputs:
+      package:
```
jl-wynen, line 6: Does it make sense to use a `choice` type here (https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputs) That way, the caller can't make a typo. But we would have to maintain the list as we add packages.

.github/workflows/docs-deploy.yml
```diff
@@ -2,56 +2,55 @@ name: Deploy docs
...
-    permissions:
-      contents: write
-    steps:
-      - uses: actions/checkout@v6
-      - uses: actions/download-artifact@v4
-        with:
-          path: site
-          pattern: docs-*
-          merge-multiple: false
+        run: pixi run -e docs-${{ inputs.package }} docs ${{ inputs.package }}
```
jl-wynen:
```suggestion
        run: pixi run docs-${{ inputs.package }}}}
```

.github/workflows/docs-deploy.yml
jl-wynen, line 1: We have 3 workflows that build docs, 2 of them also publish the docs to different places. Can we merge these into a reusable workflow and call that from CI, releases, and the web UI?

.github/workflows/docs-deploy.yml
nvaytet: We had a list before, so I think it would be ok to have a choice here? especially if it translates to a dropdown menu in the web ui?

.github/workflows/docs-deploy.yml
nvaytet: Yes, that's what I felt when I was trying to fix this.
But I felt it would be complicated to get right first time, and I did not want this to block releases for the other packages.
So we can fix as a follow-up?

nvaytet:
Note that CI fails because I manually stopped it, wasn't worth using up more CI time for the latest commit.

.github/workflows/docs-deploy.yml
jl-wynen: Ok. Can you open an issue?

jl-wynen (approved):

PR #263, opened 2026-03-24
Prepare imaging and nmx releases

nvaytet:
Do I need to do anything else than bump the essreduce version in the pyproject files?


YooSunYoung:
> Do I need to do anything else than bump the essreduce version in the pyproject files?

I guess that's it...?

YooSunYoung (approved):

PR #264, opened 2026-03-24
[Essnmx] Use time bin width instead of number of bin edges.

YooSunYoung:
Requested by @aaronfinke .

Now you can set `--time-bin-width` instead of `nbins`.
I didn't remove the `nbins` option so you can still use it, but you need to set the `--time-bin-width 0`.

## Command Example
```
essnmx-reduce --input-file nmx.hdf --time-bin-width 3 --time-bin-unit ms
# Or
essnmx-reduce --input-file nmx.hdf --time-bin-width 0 --nbins 50
```

In this case below, the `nbins` will be ignored and default `time-bin-width` of `3[ms]` will be used.
```bash
essnmx-reduce --input-file nmx.hdf --nbins 50
```

## Command Helper
```bash
essnmx-reduce -h

Workflow Configuration:
  --time-bin-coordinate {event_time_offset,time_of_flight}
                        Coordinate to bin the time data. Selecting `event_time_offset` means reduction steps are skipped, i.e. calculating `time of flight(tof)` and simply saves histograms of the raw data. (default: time_of_flight)
  --time-bin-width TIME_BIN_WIDTH
                        Width(Length) of each Time Bin in [time_bin_unit]. If `time_bin_width` and `nbins` are both given, `time_bin_width` will be preferred. Set it to `0` if you want to use `nbins` instead. (default: 3)
  --nbins NBINS         Number of Time bins. If `bin_width` is given, `nbins` will be ignored. (default: 50)
```

### DIFF
--- a/packages/essnmx/src/ess/nmx/executables.py
+++ b/packages/essnmx/src/ess/nmx/executables.py
@@ -181,7 +181,7 @@ def _build_time_bin_edges(
             dim=t_coord_name, start=min_t, stop=max_t, step=time_bin_width
         )
         # If the last bin edge is smaller than `max_t`
-        if bin_edges[t_coord_name, -1] < max_t:
+        if bin_edges[t_coord_name, -1] <= max_t:
             # Need to append one more edge to cover the whole range.
             true_last_bin_edge = bin_edges[t_coord_name, -1] + time_bin_width
             bin_edges = sc.concat([bin_edges, true_last_bin_edge], dim=t_coord_name)
--- a/packages/essnmx/src/ess/nmx/executables.py
+++ b/packages/essnmx/src/ess/nmx/executables.py
@@ -23,7 +23,6 @@ from .configurations import (
     ReductionConfig,
     TimeBinCoordinate,
     WorkflowConfig,
-    validate_time_bin_config,
 )
 from .nexus import (
     _check_file,
@@ -170,14 +169,20 @@ def _build_time_bin_edges(
         )
 
     # If either min/max were manually selected and bin width is set.
-    if wf_config.time_bin_width > 0:
+    if wf_config.nbins is None:
+        if wf_config.time_bin_width is None:
+            time_bin_width = sc.scalar(3, unit='ms').to(unit=wf_config.time_bin_unit)
+        elif wf_config.time_bin_width is not None:
+            time_bin_width = sc.scalar(
+                wf_config.time_bin_width, unit=wf_config.time_bin_unit
+            )
         # We do not return a scalar bin width since we histogram
         # detector panel individually.
         return sc.arange(
             dim=t_coord_name,
             start=min_t.to(unit=wf_config.time_bin_unit),
             stop=max_t.to(unit=wf_config.time_bin_unit),
-            step=sc.scalar(wf_config.time_bin_width, unit=wf_config.time_bin_unit),
+            step=time_bin_width,
         )
     else:  # Number of bin edges are given but not the bin width.
         n_edges = wf_config.nbins + 1
@@ -222,8 +227,6 @@ def reduction(
     # Check the file output configuration before we start heavy computation.
     if not config.output.skip_file_output:
         _check_file(config.output.output_file, config.output.overwrite)
-
-    validate_time_bin_config(config=config)
 
     display = _retrieve_display(logger, display)
     input_file_path = _retrieve_input_file(config.inputs.input_file).resolve()
--- a/packages/essnmx/src/ess/nmx/executables.py
+++ b/packages/essnmx/src/ess/nmx/executables.py
@@ -193,7 +193,7 @@ def _build_time_bin_edges(
         if min_t.unit != max_t.unit:
             min_t = min_t.to(unit=wf_config.time_bin_unit)
             max_t = max_t.to(unit=wf_config.time_bin_unit)
-        
+
         # Avoid dropping the event that has the exact same
         # `event_time_offset`` or `tof` value as the upper bin edge.
         max_t.value = np.nextafter(max_t.value, np.inf)
--- a/packages/essnmx/src/ess/nmx/executables.py
+++ b/packages/essnmx/src/ess/nmx/executables.py
@@ -23,6 +23,7 @@ from .configurations import (
     ReductionConfig,
     TimeBinCoordinate,
     WorkflowConfig,
+    validate_time_bin_config,
 )
 from .nexus import (
     _check_file,
@@ -168,9 +169,22 @@ def _build_time_bin_edges(
             "Please check your configurations again."
         )
 
-    # Build the bin-edges to histogram the results.
-    n_edges = wf_config.nbins + 1
-    return sc.linspace(dim=t_coord_name, start=min_t, stop=max_t, num=n_edges)
+    # If either min/max were manually selected and bin width is set.
+    if wf_config.time_bin_width > 0:
+        # We do not return a scalar bin width since we histogram
+        # detector panel individually.
+        return sc.arange(
+            dim=t_coord_name,
+            start=min_t.to(unit=wf_config.time_bin_unit),
+            stop=max_t.to(unit=wf_config.time_bin_unit),
+            step=sc.scalar(wf_config.time_bin_width, unit=wf_config.time_bin_unit),
+        )
+    else:  # Number of bin edges are given but not the bin width.
+        n_edges = wf_config.nbins + 1
+        if min_t.unit != max_t.unit:
+            min_t = min_t.to(unit=wf_config.time_bin_unit)
+            max_t = max_t.to(unit=wf_config.time_bin_unit)
+        return sc.linspace(dim=t_coord_name, start=min_t, stop=max_t, num=n_edges)
 
 
 def reduction(
@@ -208,6 +222,8 @@ def reduction(
     # Check the file output configuration before we start heavy computation.
     if not config.output.skip_file_output:
         _check_file(config.output.output_file, config.output.overwrite)
+
+    validate_time_bin_config(config=config)
 
     display = _retrieve_display(logger, display)
     input_file_path = _retrieve_input_file(config.inputs.input_file).resolve()
@@ -253,22 +269,23 @@ def reduction(
         wf_config=config.workflow, result_das=tof_das, t_coord_name=t_coord_name
     )
 
+    # Histogram detector counts
+    tof_histograms = sc.DataGroup()
+    for detector_name, tof_da in tof_das.items():
+        t_coord_unit = tof_da.bins.coords[t_coord_name].unit
+        histogram = tof_da.hist({t_coord_name: t_bin_edges.to(unit=t_coord_unit)})
+        tof_histograms[detector_name] = histogram
+
+    _tof_histogram = next(iter(tof_histograms.values()))
     monitor_metadata = NMXMonitorMetadata(
         tof_bin_coord=t_coord_name,
         # TODO: Use real monitor data
         # Currently NMX simulations or experiments do not have monitors
         data=sc.DataArray(
-            coords={t_coord_name: t_bin_edges},
-            data=sc.ones_like(t_bin_edges[:-1]),
+            coords={t_coord_name: _tof_histogram.coords[t_coord_name]},
+            data=sc.ones_like(_tof_histogram.data),
         ),
     )
-
-    # Histogram detector counts
-    tof_histograms = sc.DataGroup()
-    for detector_name, tof_da in tof_das.items():
-        histogram = tof_da.hist({t_coord_name: t_bin_edges})
-        tof_histograms[detector_name] = histogram
-
     detector_results = sc.DataGroup(
         {
             detector_name: NMXReducedDetector(
--- a/packages/essnmx/src/ess/nmx/executables.py
+++ b/packages/essnmx/src/ess/nmx/executables.py
@@ -179,12 +179,21 @@ def _build_time_bin_edges(
         # We do not return a scalar bin width since we histogram
         # detector panels individually
         # and all histograms should have the same bin edges.
-        return sc.arange(
-            dim=t_coord_name,
-            start=min_t.to(unit=wf_config.time_bin_unit),
-            stop=max_t.to(unit=wf_config.time_bin_unit),
-            step=time_bin_width,
+        min_t = min_t.to(unit=wf_config.time_bin_unit)
+        max_t = max_t.to(unit=wf_config.time_bin_unit)
+        bin_edges = sc.arange(
+            dim=t_coord_name, start=min_t, stop=max_t, step=time_bin_width
         )
+        # Check if the last bin edge is equal to the max_t
+        if bin_edges[t_coord_name, -1] == max_t:
+            ...  # Perfect!
+        else:
+            # Need to append one more edge to cover the whole range.
+            true_last_bin_edge = bin_edges[t_coord_name, -1] + time_bin_width
+            bin_edges = sc.concat([bin_edges, true_last_bin_edge], dim=t_coord_name)
+
+        return bin_edges
+
     else:  # Number of bin edges are given but not the bin width.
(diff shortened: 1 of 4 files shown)


aaronfinke:
`--time-bin-width` and `--nbins` should be mutually exclusive options (e.g. use `ArgumentParser.add_mutually_exclusive_group()` )

The essnmx-reduce output should include the number of generated bins and also (this would be nice) number of events per bin.

YooSunYoung:
The failing tests seem to be related to the resource issue....?? It doesn't fail on my workstation... I'll have a look at it tmr...

YooSunYoung:
@aaronfinke 

> --time-bin-width and --nbins should be mutually exclusive options (e.g. use ArgumentParser.add_mutually_exclusive_group() )

Okay...!

> The essnmx-reduce output should include the number of generated bins and also (this would be nice) number of events per bin.

You mean the logging/printing them out?
Or you want an additional 1-D histogram in the output file?

aaronfinke:
@YooSunYoung 

> You mean the logging/printing them out?
> Or you want an additional 1-D histogram in the output file?

Certainly as logger output, yes, but also a cute 1D histogram would also be useful. `dials.find_spots` does something similar in its output:
```
Histogram of per-image spot count for imageset 0:
1879 spots found on 49 images (max 78 / bin)
                     *                           
                     *                           
                    **                          *
               **   ***              *          *
      *   ** * *** ******   * * **   *          *
      ****** ************* ** * **   ***        *
     ******* ****************** **   ***   **** *
******************************* **** *** ********
*************************************************
*************************************************
1                    image                     49
```

aaronfinke:
DIALS code for making that histogram:  

https://github.com/dials/dials/blob/main/src/dials/util/ascii_art.py

YooSunYoung:
@aaronfinke 
Now it's not allowed to give both `nbins` and `time-bin-width`. I'll add the 1D histogram too (#267).

```bash
pixi run essnmx-reduce --nbins 3 --time-bin-width 3
usage: essnmx-reduce [-h] --input-file INPUT_FILE [INPUT_FILE ...] [--swmr] [--detector-ids DETECTOR_IDS [DETECTOR_IDS ...]] [--iter-chunk] [--chunk-size-pulse CHUNK_SIZE_PULSE] [--chunk-size-events CHUNK_SIZE_EVENTS]
                     [--time-bin-coordinate {event_time_offset,time_of_flight}] [--min-time-bin MIN_TIME_BIN] [--max-time-bin MAX_TIME_BIN] [--time-bin-unit {ms,us,ns}] [--lookup-table-file-path LOOKUP_TABLE_FILE_PATH]
                     [--tof-simulation-num-neutrons TOF_SIMULATION_NUM_NEUTRONS] [--tof-simulation-min-wavelength TOF_SIMULATION_MIN_WAVELENGTH] [--tof-simulation-max-wavelength TOF_SIMULATION_MAX_WAVELENGTH]
                     [--tof-simulation-min-ltotal TOF_SIMULATION_MIN_LTOTAL] [--tof-simulation-max-ltotal TOF_SIMULATION_MAX_LTOTAL] [--tof-simulation-seed TOF_SIMULATION_SEED] [--time-bin-width TIME_BIN_WIDTH | --nbins NBINS] [--verbose] [--skip-file-output]
                     [--output-file OUTPUT_FILE] [--overwrite] [--compression {NONE,GZIP,BITSHUFFLE_LZ4}]
essnmx-reduce: error: argument --time-bin-width: not allowed with argument --nbins
```

YooSunYoung:
I just realized... we probably want to set the bid width in angstrom

aaronfinke:
> I just realized... we probably want to set the bid width in angstrom

Not necessarily. Setting by TOF is more intuitive. I would keep it as seconds or milliseconds.

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -168,9 +168,28 @@ def _build_time_bin_edges(
...
         )
 
-    # Build the bin-edges to histogram the results.
-    n_edges = wf_config.nbins + 1
-    return sc.linspace(dim=t_coord_name, start=min_t, stop=max_t, num=n_edges)
+    # If either min/max were manually selected and bin width is set.
+    if wf_config.nbins is None:
+        if wf_config.time_bin_width is None:
+            time_bin_width = sc.scalar(3, unit='ms').to(unit=wf_config.time_bin_unit)
+        elif wf_config.time_bin_width is not None:
```
nvaytet:
```suggestion
        else:
```
?

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -168,9 +168,28 @@ def _build_time_bin_edges(
...
+            time_bin_width = sc.scalar(
+                wf_config.time_bin_width, unit=wf_config.time_bin_unit
+            )
+        # We do not return a scalar bin width since we histogram
+        # detector panel individually.
```
nvaytet:
```suggestion
        # detector panels individually.
```

nvaytet: Also, I did not entirely understand the comment.
Is it saying that we need to return the same edges for all panels, and if only a single scalar is given, when histogramming scipp will find different limits for each panel and would result in different edges per panel?

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -168,9 +168,28 @@ def _build_time_bin_edges(
...
+        return sc.arange(
+            dim=t_coord_name,
+            start=min_t.to(unit=wf_config.time_bin_unit),
+            stop=max_t.to(unit=wf_config.time_bin_unit),
+            step=time_bin_width,
```
nvaytet: I think we need to be careful with using `arange` here. Depending on the step size, you can be dropping the max edge.
For example, `np.arange(1, 17.5, 3)` yields `array([ 1.,  4.,  7., 10., 13., 16.])`. So the edges would stop at 16, missing everything after that.
Similarly, it could be surprising that `np.arange(2, 18., 2)` stops at 16, while `np.arange(2, 18.01, 2)` stops at 18...

packages/essnmx/tests/executable_test.py
```diff
@@ -346,6 +430,16 @@ def lut_file_path(tmp_path: pathlib.Path):
 def test_reduction_with_lut_file(
     reduction_config: ReductionConfig, lut_file_path: pathlib.Path
 ) -> None:
+    # We cannot use `time_bin_width` configuration
+    # since the lookup table is written in wavelength
+    # and it is not trivial to change lookup table
+    # and test the expected time of flight result by setting time bin width.
+    # i.e. if the look up table wavelength rage changes,
```
nvaytet:
```suggestion
    # i.e. if the lookup table wavelength range changes,
```

packages/essnmx/tests/executable_test.py
```diff
@@ -346,6 +430,16 @@ def lut_file_path(tmp_path: pathlib.Path):
...
+    # the number of bins changes and the histogram data sizes changes.
+    # This test is only for checking if the look up table is used as expected or not
+    # therefore using number of bins should be fine.
```
nvaytet, line 440: Not sure I understood why changes in the resolution/range in the lookup table would cause changes in number of bins in the reduced result? What did I miss?

packages/essnmx/tests/executable_test.py
```diff
@@ -69,7 +69,7 @@ def _check_non_default_config(testing_config: ReductionConfig) -> None:
         testing_model = testing_child.model_dump(mode='python')
         default_model = default_child.model_dump(mode='python')
         for key, testing_value in testing_model.items():
-            if key == 'lookup_table_file_path':
+            if key in ['lookup_table_file_path', 'nbins']:
```
nvaytet: Shouldn't this also include `time_bin_width`? It can also be `None`, right?

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -253,22 +272,23 @@ def reduction(
...
     )
 
+    # Histogram detector counts
+    tof_histograms = sc.DataGroup()
+    for detector_name, tof_da in tof_das.items():
+        t_coord_unit = tof_da.bins.coords[t_coord_name].unit
+        histogram = tof_da.hist({t_coord_name: t_bin_edges.to(unit=t_coord_unit)})
+        tof_histograms[detector_name] = histogram
+
+    _tof_histogram = next(iter(tof_histograms.values()))
```
nvaytet, line 290: A bit annoying that we have to do this, instead of just using the `t_bin_edges` below.
I understand it's because of
```Py
histogram = tof_da.hist({t_coord_name: t_bin_edges.to(unit=t_coord_unit)})
```
above.

Does that mean that the `t_coord_unit` can be different for different detector panels? How likely is that?

Is the unit of the tof coordinate in the reduced data something the user can control via one of the args?
If so, maybe the provider that computes the tof should do the conversion to that unit, and then we would know the unit from the argument and could avoid some gymnastics here?

Being able to say whether you want the output in microseconds or milliseconds sounds like something the users might want?

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: Yes...! Should I elaborate more...? like...

```suggestion
        # detector panels individually
        # and all histograms should have the same bin edges.
```

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: Hmm yeah... we might not want to lose the last bin... should I just calculate the number of bins and correct last edge, and use linspace instead?

nvaytet: Well we / @aaronfinke need to decide which is the quantity that should be preserved.
If we specify a range with min and max, and a bin width, but the range does not span an integer number of bin widths, should we:

1. preserve the bin width and adjust the start or end of the range with some padding to allow an integer number of bin widths
2. preserve the range and compute a number of bins based on the requested width, use linspace, and end up with a bin width which will not exactly match the requested width
3. Something else?

packages/essnmx/tests/executable_test.py
YooSunYoung:
```suggestion
            if key in ['lookup_table_file_path', 'nbins', 'time_bin_width']:
```

packages/essnmx/tests/executable_test.py
YooSunYoung: Because the wavelength range of reduced data changes if the lookup table wavelength range changes.
If the wavelength range shrinks, the number of bins should be smaller.

packages/essnmx/src/ess/nmx/executables.py
aaronfinke: My immediate reaction is that setting the bin length to the user defined value is more important than ensuring all bin sizes are equal, so I'd go for option 1. So setting the nbins as $floor(\Delta_{TOF}/t_{bin})$ and setting bin[0] and bin[-1] size as $t_{bin}$ + $(\Delta_{TOF}$ $modulo$ $t_{bin})/2$ makes the most sense.

YooSunYoung: We just talked offline and the conclusion was to keep the bin-width as the user-input and add an extra padding bin if needed.

I'll fix this in this way and ask for a review again.

Image

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: > Does that mean that the t_coord_unit can be different for different detector panels? How likely is that?

No... it was just me being lazy to do it only once...

But in theory they could have different units I guess. Although we probably want same units in the output.

> Being able to say whether you want the output in microseconds or milliseconds sounds like something the users might want?

Probably...? Then it's easy for us to decide the unit here.

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -168,9 +164,40 @@ def _build_time_bin_edges(
...
+        min_t = min_t.to(unit=wf_config.time_bin_unit)
+        max_t = max_t.to(unit=wf_config.time_bin_unit)
+        bin_edges = sc.arange(
+            dim=t_coord_name, start=min_t, stop=max_t, step=time_bin_width
+        )
+        # If the last bin edge is smaller than `max_t`
+        if bin_edges[t_coord_name, -1] < max_t:
+            # Need to append one more edge to cover the whole range.
+            true_last_bin_edge = bin_edges[t_coord_name, -1] + time_bin_width
+            bin_edges = sc.concat([bin_edges, true_last_bin_edge], dim=t_coord_name)
```
YooSunYoung: This was easier to read/write than calculating linspace nbins start stop...

packages/essnmx/tests/executable_test.py
nvaytet: Can you explain in what case does the wavelength range in the lookup table changes?

The lut has dimensions `(event_time_offset, Ltotal)`. The wavelength is the z (or color) values.
We do not control/constrain the range of wavelengths, that is given by the results of the `tof` simulation.
Unless you are constraining the range of wavelengths for neutrons produced in the `tof` source in the simulation? (Using `Source(wmin=1 * angstrom, wmax=5 * angstrom)`)

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -168,9 +164,40 @@ def _build_time_bin_edges(
...
+        # We do not return a scalar bin width since we histogram
+        # detector panels individually
+        # and all histograms should have the same bin edges.
+        min_t = min_t.to(unit=wf_config.time_bin_unit)
+        max_t = max_t.to(unit=wf_config.time_bin_unit)
+        bin_edges = sc.arange(
+            dim=t_coord_name, start=min_t, stop=max_t, step=time_bin_width
+        )
+        # If the last bin edge is smaller than `max_t`
+        if bin_edges[t_coord_name, -1] < max_t:
```
nvaytet:
```suggestion
        if bin_edges[t_coord_name, -1] <= max_t:
```
Either the above or `max_t` should use `nextafter`, if not we would be dropping the last event again?

packages/essnmx/src/ess/nmx/executables.py
nvaytet: I think it should be that if the user requests a bin width of 0.3 ms, then the output unit should be in ms.
So we should convert the units in the different panels to ms, then use the edges to histogram.
Does that make sense?

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: If we include the `max_t+np.nextafter`,  it'll change the bin width so... this is better I guess...?

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: I'm not sure.... user might want to pass the argument in `ms` cause it's easy but not necessarily want the output to be written in `ms`...?

I think it's not that simple. Should we fix it in another issue/PR?

packages/essnmx/tests/executable_test.py
YooSunYoung: No I was talking about when I directly modified the lookup table, not modifying the lookup table generating settings.

So that I could see the expected difference between the workflow that generate lookup table and that uses the existing lookup table.

packages/essnmx/src/ess/nmx/executables.py
nvaytet: I think we should go with that rule. I think it makes sense, and as long as we document the behaviour, it should be fine.

But I'm ok with delaying to a later PR if the changes here are needed elsewhere.

Generally, I don't think we should make interfaces too flexible, with many different buttons that can be turned. Allowing things like banks with different time units is an example: this will basically never happen, so I don't think we should make the code more complex just because of that eventuality?
If you really want to protect against it, you can add a check and raise an error if detected?

nvaytet (approved):

PR #272, opened 2026-03-26
Refactor CI failure condition logic

YooSunYoung:
I found out that skipped and success and failure are not the only conditions...


YooSunYoung:
I cancelled the test on purpose to see if it actually works...!

nvaytet (approved):

PR #274, opened 2026-04-13
fix: handle binned data in max res finder

### LINKED ISSUE #273: [essimaging] `maximum_resolution_achievable` should not hardcode "time" as the "third" dimension
Currently one of the arguments to `maximum_resolution_achievable` function (https://scipp.github.io/essimaging/generated/modules/ess.imaging.tools.resolution.maximum_resolution_achievable.html#ess.imaging.tools.resolution.maximum_resolution_achievable) is the bin-edges of the `time` coordinate.

This is not very flexible, sometimes the third coordinate is `wavelength` or `event_time_offset` or maybe something different.

To fix this we should rename the argument and update the documentation, and maybe the implementation, to make it clear the third argument can be a non-time dimension.

jokasimr:
Fixes #273 
and allows binned data in the max resolution finder function.

### DIFF
--- a/packages/essimaging/src/ess/imaging/tools/resolution.py
+++ b/packages/essimaging/src/ess/imaging/tools/resolution.py
@@ -1,3 +1,5 @@
+from math import floor, log2
+
 import numpy as np
 import scipp as sc
 from numpy.typing import NDArray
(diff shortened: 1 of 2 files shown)


packages/essimaging/src/ess/imaging/tools/resolution.py
```diff
@@ -59,7 +56,17 @@ def maximum_resolution_achievable(
...
+    events = events.copy(deep=False)
+
+    if events.bins is not None:
+        image_dims = (coarse_x_bin_edges.dim, coarse_y_bin_edges.dim)
+        for c in image_dims:
+            if c not in events.bins.coords:
+                events.bins.coords[c] = sc.bins_like(
+                    events, sc.midpoints(events.coords[c])
+                )
+        events = events.bins.concat(image_dims)
```
YooSunYoung: Why do we have to concatenate it? Is it faster if we do?
```suggestion
```

packages/essimaging/src/ess/imaging/tools/resolution.py
YooSunYoung, line 98: Sneaky typo bug......

packages/essimaging/tests/imaging/tools/maximum_resolution_monitor_test.py
```diff
@@ -70,3 +68,42 @@ def test_finds_maximum_resolution_random(seed):
...
+        events,
+        sc.linspace('x', 0, 1, 2),
+        sc.linspace('y', 0, 1, 2),
+        # Need enough tries to be sure we find the optimum
+        max_tries=100,
+    )
+
+    events.bins.coords['x'] = sc.bins_like(events, sc.midpoints(events.coords['x']))
+    events.bins.coords['y'] = sc.bins_like(events, sc.midpoints(events.coords['y']))
+    events = events.bins.concat(['x', 'y'])
```
YooSunYoung:
```suggestion
```

Here too. I think putting the event coordinate should be enough.

packages/essimaging/tests/imaging/tools/maximum_resolution_monitor_test.py
```diff
@@ -70,3 +68,42 @@ def test_finds_maximum_resolution_random(seed):
...
+        sc.ones(dims=['events'], shape=(n,)),
+        coords={
+            'x': sc.array(dims=['events'], values=np.random.random(n)),
+            'y': sc.array(dims=['events'], values=np.random.random(n)),
+            't': sc.array(dims=['events'], values=np.random.random(n)),
+        },
+    )
+    events = events.bin(x=100, y=100, t=500)
+    del events.bins.coords['x']
+    del events.bins.coords['y']
```
YooSunYoung: Is it because they are usually dropped by reduction workflows?

YooSunYoung (approved):
Looks good to me...! Just some minor questions/suggestions.

packages/essimaging/src/ess/imaging/tools/resolution.py
jokasimr: The problem is that if we don't concatenate the bins then when we bin it again with new bin-edges in x and y we might run out of memory because of https://github.com/scipp/scipp/issues/3872.

Does that answer the question?

packages/essimaging/tests/imaging/tools/maximum_resolution_monitor_test.py
jokasimr: Yes, it's to make it look like typical detector event data in our workflows.

Typically we don't have pixel positions on the event data.

packages/essimaging/src/ess/imaging/tools/resolution.py
YooSunYoung: Yes it does...!
So it'll be not necessary if the bug is fixed right...?

Neverthless, I'm okay with it. It shoudn't be blocked by the bug fix...

packages/essimaging/tests/imaging/tools/maximum_resolution_monitor_test.py
jokasimr: Same issue here, if we bin again in `x` and `y` when there are already `x` and `y` dimensions then we might run out of memory.

packages/essimaging/src/ess/imaging/tools/resolution.py
jokasimr: Yes if the bug is fixed we should be fine without it.

YooSunYoung (approved):

jokasimr:
Although, now that I'm thinking more about it, we don't typically have 'x' and 'y' coords on our detector data either :thinking: and this method currently requires that. There might be a better / simpler way to do it that does not require any coordinates on the input data.

jokasimr:
The old implementation had some issues.

1. It required an `x` and an `` coordinate on the input data, something we typically don't have. We typically have a `position`, but it's not a dimension coordinate because it depends on both `x` and `y`, and that makes it difficult to work with in this context. The user could of course always add their own coordinates, but that is cumbersome.
2. To be able to re-bin the event data on different pixel grids it had to add "aux" coordinates on the binned data array to keep track of what pixel each event belonged to originally. This is both quite inefficient for large event data, and it required either to modify the event data, or it would have required a copy, neither of which are not good options.
3. If we bin into a number of pixels in each dimension that does not divide the original number of pixels then more of the "original" pixels will fall into some of the new pixels ("aliasing"). This isn't a big problem, because the effect is canceled in the normalization, but it can look ugly / surprising, so it's probably better to make sure that each "coarse" pixel contains the same number of "fine" pixels.

These issues are solved if we "fold" the original fine grid to coarser grids instead of re-binning the events, so that is the approach taken in the new implementation.

YooSunYoung:
>  It required an x and an `` coordinate on the input data, something we typically don't have. We typically have a position, but it's not a dimension coordinate because it depends on both `x` and `y`, and that makes it difficult to work with in this context. The user could of course always add their own coordinates, but that is cumbersome.

One thing to be clear here:
We can't use the real position x and y of positions. (assuming that it is calculated from the NXtransformation chain.)
It should be `x_pixel_offset` and `y_pixel_offset` that we use as `x` and `y`.

It's usually easy to fold the data along `x_pixel_offset` and `y_pixel_offset` if the events are grouped by pixel ids but it's not guaranteed that we can do that.
We didn't think about it because we were quite sure that timepix and orca camera has the pixel ids (detector_numbers) in a way that we can easily fold into x and y.

But I think the new signature is much better...!

PR #276, opened 2026-04-14
[POLARIZATION] Add esspolarization

jokasimr:


nvaytet:
I transfered the issues over from `esspolarization`

jokasimr:
> I transfered the issues over from `esspolarization`

Thanks!

packages/esspolarization/CODE_OF_CONDUCT.md
nvaytet, line 1: I guess like in the essreflectometry PR, this file as well as `CONTRIBUTING.md` and `LICENSE` should be removed.

packages/esspolarization/pyproject.toml
```diff
@@ -0,0 +1,135 @@
...
+    "E501",  # longer lines are sometimes more readable
+    "F403",  # *-imports used with domain types
+    "F405",  # linter may fail to find names because of *-imports
+    "I",  # we don't collect imports at the top
+    "S101",  # asserts are used for demonstration and are safe in notebooks
+    "T201",  # printing is ok for demonstration purposes
+]
+
+[tool.ruff.format]
+quote-style = "preserve"
```
nvaytet:
```suggestion
```

nvaytet (commented):
We also need to update the main README file (table and dependency graph)

nvaytet:
Question: should esspolarization just be a sub module of essreduce?

I think essreduce did not exist when we first created it, but maybe we can move it in there?

jokasimr:
> Question: should esspolarization just be a sub module of essreduce?

That makes sense to me.
What do you think @SimonHeybrock?

SimonHeybrock:
Might make sense, one package fewer?

nvaytet:
So do we still want to keep these changes or just directly move it into `essreduce`?

jokasimr:
I'm fine with directly moving it into essreduce.

nvaytet:
Ok, then let's do that instead.
I don't think we currently have any packages that depend on `esspolarization`, so it should be fine?

jokasimr:
> Ok, then let's do that instead.
I don't think we currently have any packages that depend on esspolarization, so it should be fine?

Essreflectometry does, but it's fine anyway, even if we move the polarization logic to essreduce we don't need to update essreflectometry at the same time.

YooSunYoung (approved):

jokasimr:
@YooSunYoung This is fine right? Now the only changes are in `packages/esspolarization` (and in `pixi.lock` but it's not important).

PR #434, opened 2026-04-17
[ESSREDUCE] Move `esspolarization` to `essreduce`

jokasimr:
As suggested in #276


packages/essreduce/src/ess/reduce/polarization/__init__.py
```diff
@@ -0,0 +1,90 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+# ruff: noqa: E402, F401, I
```
nvaytet:
```suggestion
```
Not sure this is needed?

packages/essreduce/src/ess/reduce/polarization/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+
+import importlib.metadata
+
+try:
+    __version__ = importlib.metadata.version("esspolarization")
+except importlib.metadata.PackageNotFoundError:
+    __version__ = "0.0.0"
+
+del importlib
```
nvaytet: I guess this should go away? It should just use the version of `essreduce`.

packages/essreduce/src/ess/reduce/polarization/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+
+
```
nvaytet, line 8: Maybe we can add a docstring here for the `polarization` submodule that describes a little the contents?

packages/essreduce/docs/api-reference/index.md
```diff
@@ -31,6 +31,7 @@
    logging
```
nvaytet, line 31: I think you are missing the docs contents of `esspolarization`? e.g. https://scipp.github.io/esspolarization/user-guide/sans-polarization-analysis-methodology.html

That said, I don't know how much of it we want to keep, but I would say at least to start we would want to keep all of it?

packages/essreduce/src/ess/reduce/polarization/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+    "Polarizer",
+    "PolarizingElement",
+    "ReducedSampleDataBySpinChannel",
+    "SecondDegreePolynomialEfficiency",
+    "SupermirrorEfficiencyFunction",
+    "SupermirrorWorkflow",
+    "TotalPolarizationCorrectedData",
+    "TransmissionFunction",
+    "Up",
+]
```
nvaytet, line 85: I think you should also add `polarization` in the top level [init](https://github.com/scipp/ess/blob/main/packages/essreduce/src/ess/reduce/__init__.py#L6) ?

packages/essreduce/docs/api-reference/index.md
jokasimr: Yes that's true, I forgot to move that :+1:

jokasimr:
Although... Doing it this way we loose the history of the ~essreduce~esspolarization package. That's not good. Maybe it's better to merge #276 instead and then do the move after?

YooSunYoung:
> Although... Doing it this way we loose the history of the essreduce package. That's not good. Maybe it's better to merge https://github.com/scipp/ess/pull/276 instead and then do the move after?

You mean the history of the esspolarization? 
I vote for merging it first and then move... 
then it's more clear who contributed to the code and that we moved the package into the essreduce.

In that case, in #276, we don't need to update pixi.toml or readme since we know that it'll be merged into essreduce.

nvaytet:
> Maybe it's better to merge https://github.com/scipp/ess/pull/276 instead and then do the move after?

Yes! do that :slightly_smiling_face:

PR #437, opened 2026-04-20
Add options for essdiffraction, esspolarization, and essreflectometry…

MridulS:
… in manual docs deploy


.github/workflows/docs-deploy.yml
```diff
@@ -11,6 +11,9 @@ on:
           - essreduce
           - essimaging
           - essnmx
+          - essdiffraction
+          - esspolarization
```
nvaytet:
```suggestion
```

We know that this will be moved into `essreduce`, so no need to add it here.

YooSunYoung (approved):

PR #447, opened 2026-04-22
Do not run CI when merging to main

nvaytet:
Do we really need to run the CI again when we merge into `main`?

The tests have already run in the PRs.
Requiring branches to be up to date with `main` or using the merge queue should make sure we dont merge stuff that is out of date…
It would save quite a lot of CI jobs?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
               - 'pyproject.toml'
               - 'pixi.lock'
               - 'pixi.toml'
-              - '.github/workflows/**'
+              - '.github/workflows/ci.yml'
               - '.pre-commit-config.yaml'
               - '.python-version'
             essimaging:
@@ -38,7 +38,7 @@ jobs:
               - 'pyproject.toml'
               - 'pixi.lock'
               - 'pixi.toml'
-              - '.github/workflows/**'
+              - '.github/workflows/ci.yml'
               - '.pre-commit-config.yaml'
               - '.python-version'
             essnmx:
@@ -47,7 +47,7 @@ jobs:
               - 'pyproject.toml'
               - 'pixi.lock'
               - 'pixi.toml'
-              - '.github/workflows/**'
+              - '.github/workflows/ci.yml'
               - '.pre-commit-config.yaml'
               - '.python-version'
             essreflectometry:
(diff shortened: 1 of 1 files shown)


YooSunYoung:
Using the merge queue seems to imply that we can get rid of the CI running on main.


But why do we require PRs to be up to date with main though...?
We will often have a lot of parallel works in subpackages that don't affect each other.

nvaytet:
> But why do we require PRs to be up to date with main though...?

I am not sure we need to. If I understood the docs correctly, I think the merge queue takes care of that?

YooSunYoung:
Yeah it looks like we can't skip the queue when we have one... I'm okay with this change then.

YooSunYoung (approved):

.github/workflows/ci.yml
```diff
@@ -142,7 +140,7 @@ jobs:
   report:
     name: Report Job Status
     needs: [changes, formatting, test, docs]
-    if: always() && ( github.ref != 'refs/heads/main' )  # This job is for branch protection rule so it doesn't have to run on main branch.
+    if: always()
```
nvaytet: @YooSunYoung do we still need the `if: always()` here?
or is it equivalent to not having any `if` condition?

YooSunYoung: We need it otherwise it'll not run when any of the `needs` job fails.
Let's add the comment here.

```suggestion
    if: always()  # Should always run regardless of the state of the needs jobs.
```

.github/workflows/ci.yml
YooSunYoung: Can we be more specific ...?
We don't have to run the tests when we update other CI files.

```suggestion
              - '.github/workflows/ci.yml'
```

YooSunYoung: Here too.
```suggestion
              - '.github/workflows/ci.yml'
```

YooSunYoung:
```suggestion
              - '.github/workflows/ci.yml'
```

YooSunYoung:
```suggestion
              - '.github/workflows/ci.yml'
```

YooSunYoung:
```suggestion
              - '.github/workflows/ci.yml'
```

YooSunYoung:
```suggestion
              - '.github/workflows/ci.yml'
```

YooSunYoung (changes requested):

YooSunYoung (approved):

PR #449, opened 2026-04-22
Add package logos to search results

MridulS:
This should add technique logos next to the search results on the homepage.


nvaytet:
All our logos are quite similar... I wonder if visually it would be more helpful to use icons a bit like the ones from the ESS instrument suite? https://www.sciencedirect.com/science/article/pii/S0168900220300097
Screenshot_20260422_192116

Those ones might not really work as they would span multiple technique packages, but maybe we can find something similar online or ask chatgpt to make some for us?

nvaytet:
Alternatively, make our logos more different?
But now that we are unifying our docs pages somewhat, maybe it's good that they all have the same look & feel?

nvaytet:
I asked chatgpt, it made this
ChatGPT Image Apr 23, 2026, 09_43_18 AM

nvaytet (approved):
Either add package names inside/next to logos in this PR or address it later.

PR #450, opened 2026-04-22
[ESSSANS] Rename loki monitors to use m prefix

### LINKED ISSUE #337: Update hardcoded LOKI monitor names to use m-prefix convention
## Context

LOKI monitor naming was recently updated to use an `m` prefix (e.g., `beam_monitor_1` → `beam_monitor_m1`, `beam_monitor_3` → `beam_monitor_m3`).

`LokiWorkflow()` still hardcodes the old names as defaults:

```python
# Currently in LokiWorkflow defaults:
NeXusMonitorName[Incident] = 'beam_monitor_1'
NeXusMonitorName[Transmission] = 'beam_monitor_3'
```

These should be updated to:

```python
NeXusMonitorName[Incident] = 'beam_monitor_m1'
NeXusMonitorName[Transmission] = 'beam_monitor_m3'
```

## Impact

In ESSlivedata, we now pass the resolved monitor names from the aux source configuration into the factory, overriding the defaults. So the live dashboard is not affected. However, offline users of `LokiWorkflow()` who rely on the defaults will get stale names.

nvaytet:
Had to make new files for CI with the correct monitor names.

I include the script that was used to only keep the first 5 pulses of data to make the tests run faster.

Fixes #337


packages/esssans/tests/loki/conftest.py
```diff
@@ -68,7 +68,7 @@ def make_workflow() -> sciline.Pipeline:
         # For now, use a dummy file for all runs. This produces a result with all NaNs,
         # but is sufficient to test that the workflow runs end-to-end.
         # TODO: Replace with real test data when available.
-        file = loki.data.loki_coda_file_one_event()
```
nvaytet, line 71: No need to use another file with just one event, we simply use the same file as for the detector diagnostics, it's small enough.

SimonHeybrock (approved):

PR #543, opened 2026-04-27
feat: add monitor to estia workflow

jokasimr:
Adds an optional monitor normalization step to the Estia workflow.


packages/essreflectometry/src/ess/estia/corrections.py
```diff
@@ -21,6 +23,47 @@
...
+
+    See also
+    --------
+    ess.reduce.normalization.normalize_by_monitor_histogram:
+        For details and the actual implementation.
+    """
+    return ess.reduce.normalization.normalize_by_monitor_histogram(
+        detector=detector,
+        monitor=monitor,
+        uncertainty_broadcast_mode=UncertaintyBroadcastMode.drop,
```
jl-wynen: Why did you hardcode this? Does the workflow not have this parameter?

packages/essreflectometry/src/ess/estia/corrections.py
```diff
@@ -55,6 +102,6 @@ def correct_by_footprint(da: sc.DataArray) -> sc.DataArray:
     return da / sc.sin(da.coords['theta'])
 
 
-default_corrections = {correct_by_proton_current, correct_by_footprint}
+default_corrections = {correct_by_proton_current, correct_by_footprint, 'monitor'}
```
jl-wynen: Why did you break with the established interface?

jl-wynen: Does it makes sense to normalise by both monitor and proton charge? This would first remove the dependence on the flux and then add it back in.
I think we should require the user to specify a normalisation and either do none by default or only one of the two.

packages/essreflectometry/src/ess/estia/workflow.py
```diff
@@ -80,11 +83,15 @@ def default_parameters() -> dict:
...
-        CorrectionsToApply: corrections.default_corrections,
+        CorrectionsToApply: corrections.default_corrections - {'monitor'},
         DetectorSpatialResolution: 0.0025 * sc.units.m,
         LookupTableRelativeErrorThreshold: {
             "multiblade_detector": float('inf'),
         },
+        # The monitor is missing from the Nexus files,
+        # so to be able to load Nexus files anyway
+        # this is set to None.
+        WavelengthMonitor[RunType]: None,
```
jl-wynen: How would you use the workflow with monitor norm? It looks to me like you need to either implement your own `EstiaWorkflow` to omit this parameter or manually add all monitor providers to the workflow.

In powder diffraction, we have an argument in the workflow constructor function that sets the normalisation mode: https://scipp.github.io/ess/diffraction/user-guide/dream/dream-advanced-powder-reduction.html#Normalize-by-integrated-monitor
That way, you get the appropriate graph without having to build it yourself. Plus, you don't end up with spurious dependencies on a monitor or proton charge that doesn't exist.

packages/essreflectometry/src/ess/estia/corrections.py
jokasimr: Aha maybe it does? I didn't think the workflow had the uncertainty broadcast mode parameter. I'll change it.

packages/essreflectometry/src/ess/estia/corrections.py
jokasimr: Because the monitor correction is not a function of the coordinates on the event data, like the footprint correction or the proton current correction.

It could be made into one, if we add the monitor intensity as a event coordinate on the data, like is done with the proton current. But then we can't use the helper from essreduce.

I agree it's a bit inconsistent, but I don't think we currently have another good solution for how to "specify what corrections should be applied" in the workflows. But I'm open to hear suggestions for alternatives.

jokasimr: Yes I think you are right.

packages/essreflectometry/src/ess/estia/workflow.py
jokasimr: I thought you would create the workflow using `EstiaWorkflow` and then `.insert()` the provider for `WavelengthMonitor[RunType]` . But I see now that you are right, setting the `WavelengthMonitor` parameter on the workflow removes all dependencies upstream from `WavelengthMonitor` from the workflow :/ so even if you add a provider for `WavelengthMonitor` you'd still be missing all the providers it depends on... 

I'll try to do something more like what was done in the Dream workflow. I was trying to do something simpler here.

jokasimr:
@jl-wynen was there anything else you wanted changed here?

packages/essreflectometry/src/ess/estia/workflow.py
jl-wynen: How do you create workflows now with the latest changes?

jokasimr:
```
wf = EstiaWorkflow()

# To add monitor correction
wf[CorrectionsToApply] = wf.compute(CorrectionsToApply) | {'monitor'}

# To remove monitor correction
wf[CorrectionsToApply] = wf.compute(CorrectionsToApply) - {'monitor'}
```

But it's probably better to explicitly set the corrections that you want to use, instead of modifying the default setting.

packages/essreflectometry/src/ess/estia/corrections.py
```diff
@@ -61,6 +108,6 @@ def assume_time_series_constant_with_zero_default_value_if_empty(da: sc.DataArra
     return da.mean() if len(da) > 0 else sc.scalar(0.0, unit=da.unit)
 
 
-default_corrections = {correct_by_proton_current, correct_by_footprint}
+default_corrections = {correct_by_footprint, 'proton_current'}
```
jl-wynen:
```suggestion
default_corrections = {correct_by_footprint}
```
because the default is set in the workflows without proton current correction.

packages/essreflectometry/src/ess/estia/corrections.py
```diff
@@ -30,6 +70,8 @@ def add_coords_masks_and_apply_corrections(
     bdlim: BeamDivergenceLimits,
     wbins: WavelengthBins,
     proton_current: ProtonCurrent[RunType],
+    monitor: WavelengthMonitor[RunType],
```
jl-wynen: This is a problem. You now have to always load and process both the monitor and proton charge even if you don't need them. Plus, this means that the graph visualization is not enough to understand the data flow. 
So I still think normalisations should be selected on the graph level, not as a parameter.

packages/essreflectometry/src/ess/freia/corrections.py
```diff
@@ -21,6 +19,48 @@
 )
 from .conversions import add_coords
 from .maskings import add_masks
+from .types import WavelengthMonitor
+
+
+def normalize_by_monitor_histogram(
```
jl-wynen, line 26: Do we need this copy of the estia function? Can you move this to `reflectometry.corrections`?

packages/essreflectometry/src/ess/estia/corrections.py
jokasimr: The Estia workflow should have proton current correction. Right now it's unclear how to load that information from the nexus files, but I don't think we should adjust the default parameters because of that.

packages/essreflectometry/src/ess/freia/corrections.py
jokasimr: Freia and Estia don't have the same `WavelengthMonitor` types. Freia has several monitors, while Estia only has one. I though it was less hassle to just duplicate it for now. This will definitely change when we get real requirements about how Freia wants to do monitor normalization.

jl-wynen: Then the monitor should be parametrized over the `MonitorType`. So can you replace the custom type with the one define din essreduce?
If you think it's too verbose to always specify the type when using an estia workflow, you can provide an alias for the type there. But we should avoid duplicating code just because typing `[Monitor]` is too much.

jokasimr: But that's what is already done right? I mean, there's a type alias for the essreduce monitor in `estia/types.py`.

The point is: I have no idea what the monitor normalization for Freia will look like, what monitor it will use, etc. It doesn't make any sense to do anything there right now. It's just a placeholder.

packages/essreflectometry/src/ess/estia/corrections.py
jokasimr: The idea with the `CorrectionsToApply` set was to centralize the settings for what corrections are applied in one place, that way they user has to remember fewer *things to change* to make the workflow apply the set of corrections that they have in mind.

That also unlocks some other use cases such as mapping over different sets of corrections to apply, or applying `batch_compute` over different sets of corrections to apply.

But of course it is a trade-off.
The drawbacks that you mention are: We have to always load and process the monitor and proton charge.

However, what is the cost of that?

1. Loading and processing the proton charge: Essentially zero. It's one number per pulse, the size of that dataset is measured in kB.
2. Loading and processing the monitor: Small relative to the event data. The monitor certainly has less than 1000 bins (probably significantly less, the wavelength resolution at the monitor position is poor, so there's not much point to having a dense binning), and never more than one histogram per pulse, which makes the size at most 400MB/hour. The disk read speed on VISA is ~500MB/s, so that means in the worst case scenario the additional load time is about 1s per hour of experiment.

If in some scenario this becomes a real problem, the user can disable loading the monitor by explicitly setting `WavelengthMonitor[RunType]` to `None` on the workflow.

I lean towards thinking the advantages of the usability improvements outweighing the performance drawbacks. What do you think?

jl-wynen: It's not just about loading the monitor. It also needs to be processed which includes building a lookup table and using it.

But the real issue to me is the readability of the workflow. Right now, the graph 'lies' in that it claims that the output depends on the monitor. But to know whether that really is the case, you need to look at the parameter values. Sciline workflows have no notion of optional edges or 'choice nodes'. So I can see this leading to confusion down the line when people will see the graph and wonder why they are normalising by monitor even though they disabled it.

Also, changing the corrections list changes the unit of the return value. This can impact downstream providers (the final output should stay the same). Are we sure that all downstream processing is independent of the unit?

jokasimr: I've implemented a system similar to the one that exists in DREAM now. We should probably think about if this should be generalized, because it might look the same for most instruments.

jl-wynen (approved):

PR #545, opened 2026-04-27
[NMX] Allow setting output time bin unit.

YooSunYoung:
I added output configuration to manually set output time bin unit.

At first, I thought it should be in the `output` configuration but then i realized it may be misleading when someone wants to use `save_results` function separately... So I just added it to `workflow` configuration.

### DIFF
--- a/packages/essnmx/src/ess/nmx/executables.py
+++ b/packages/essnmx/src/ess/nmx/executables.py
@@ -186,7 +186,7 @@ def _build_time_bin_edges(
             true_last_bin_edge = bin_edges[t_coord_name, -1] + time_bin_width
             bin_edges = sc.concat([bin_edges, true_last_bin_edge], dim=t_coord_name)
 
-        return bin_edges
+        return bin_edges.to(dtype=float)
 
     else:  # Number of bin edges are given but not the bin width.
         n_edges = wf_config.nbins + 1
@@ -197,7 +197,9 @@ def _build_time_bin_edges(
         # Avoid dropping the event that has the exact same
         # `event_time_offset`` or `tof` value as the upper bin edge.
         max_t.value = np.nextafter(max_t.value, np.inf)
-        return sc.linspace(dim=t_coord_name, start=min_t, stop=max_t, num=n_edges)
+        return sc.linspace(
+            dim=t_coord_name, start=min_t, stop=max_t, num=n_edges, dtype=float
+        )
 
 
 def reduction(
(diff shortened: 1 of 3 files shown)


packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -281,11 +283,14 @@ def reduction(
     )
 
     # Histogram detector counts
+    output_tunit = config.workflow.result_time_bin_unit
+    t_bin_edges = t_bin_edges.to(unit=output_tunit)
```
nvaytet, line 287: If DIALS expects `ns` (the default), does it also expect integers?
I.e. if the output unit is `ns`, do we then need to convert also to `int`?

YooSunYoung: I'm not sure... but it computes wavelength from the time again so shouldn't it prefer float...?

nvaytet: @aaronfinke ?

YooSunYoung: I asked Aaron in person but for now we want to keep the dtypes the same as the older version, which is the float64.

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -281,11 +283,14 @@ def reduction(
...
     tof_histograms = sc.DataGroup()
     for detector_name, tof_da in tof_das.items():
-        t_coord_unit = tof_da.bins.coords[t_coord_name].unit
-        histogram = tof_da.hist({t_coord_name: t_bin_edges.to(unit=t_coord_unit)})
-        tof_histograms[detector_name] = histogram
+        tof_da.bins.coords[t_coord_name] = tof_da.bins.coords[t_coord_name].to(
```
jokasimr, line 290: I guess `tof_da` is already float here?

I'm asking because if it is integer we might run into this issue if the user ever selects a larger time-unit:
```
>>> sc.scalar(10_000_000, unit='ns').to(unit='s')
<scipp.Variable> ()      int64              [s]  0
```

jokasimr (approved):
Looks good to me!

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: Yeah... that's very little likely but true...
I'll convert it to float just in case ... thanks!

```suggestion
        tof_da.bins.coords[t_coord_name] = tof_da.bins.coords[t_coord_name].to(
            dtype=float,
```

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -281,11 +283,15 @@ def reduction(
...
+            dtype=float,
+            unit=output_tunit
```
YooSunYoung:
```suggestion
            dtype=float, unit=output_tunit
```

PR #548, opened 2026-04-30
[DOCS] Update badge url and conda-forge channel web page link.

YooSunYoung:
img.shields.io now has conda/conda-channel version badge. (I think it didn't exist when we first used anaconda badges in the README...?)
Should we use them instead of anaconda badge?

(I skipped the CI btw...)

### DIFF
--- a/packages/essdiffraction/docs/conf.py
+++ b/packages/essdiffraction/docs/conf.py
@@ -203,7 +203,7 @@ html_theme_options = {
         },
         {
             "name": "Conda",
-            "url": "https://prefix.dev/channels/conda-forge/packages/essdiffraction",
+            "url": "https://anaconda.org/conda-forge/essdiffraction",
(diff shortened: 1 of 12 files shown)


packages/essdiffraction/docs/conf.py
```diff
@@ -203,7 +203,7 @@
         },
         {
             "name": "Conda",
-            "url": "https://anaconda.org/conda-forge/essdiffraction",
+            "url": "https://prefix.dev/channels/conda-forge/packages/essdiffraction",
```
jl-wynen: I would leave this as anaconda.org unless and until we make a decision to generally use prefix.dev instead. Also for installation. Do you have plans to do so? (Same thing in the rest of these changes.)

jl-wynen (commented):
I like the new badges, they are clearer about conda-forge.

jl-wynen (approved):

PR #549, opened 2026-05-01
Remove relative link to LICENSE file in pyproject.toml

MridulS:
This should fix the failing tests, earlier fix https://github.com/scipp/ess/pull/176


jokasimr:
Aha, so it was looking for the license in the wrong place, and that made it fail?

YooSunYoung (approved):

PR #555, opened 2026-05-04
[essreduce]: Allow single transformation timepoint

nvaytet:
When we provide a transformation filter that slices a single value
```Py
wf = loki.LokiWorkflow()
wf[Filename[SampleRun]] = 'loki_051657_00000135.hdf'
wf[TransformationTimeFilter[sx.NXdetector, RunType]] = lambda da: da['time', 0].data
```
we ran into the error:
```
        if isinstance(total_transform, sc.DataArray):
            time_dependent = [t for t in chain if isinstance(t, sc.DataArray)]
>           times = [da.coords['time'][0] for da in time_dependent]
E           scipp._scipp.core.DimensionError: Slicing a scalar object is not possible.
```

We fix this by returning the `.data` if the result of the `apply_transformation` is a DataArray containing just a scalar value.

### DIFF
--- a/packages/essreduce/src/ess/reduce/nexus/workflow.py
+++ b/packages/essreduce/src/ess/reduce/nexus/workflow.py
@@ -306,7 +306,12 @@ def _apply_time_filter(
 ) -> sc.Variable | sc.DataArray:
     if transform.ndim == 0 or transform.sizes == {'time': 1}:
         return transform.data.squeeze()
-    return user_filter(transform)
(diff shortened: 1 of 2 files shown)


jl-wynen:
To clarify, we got the error only when
```python
wf[TransformationTimeFilter[sx.NXdetector, RunType]] = lambda da: da['time', 0]
```
i.e., without `.data`.

jl-wynen (approved):

PR #559, opened 2026-05-10
[CI] Pin pixi version and use frozen flag to run pixi command.

YooSunYoung:
## Frozen flag
CI failed when pixi tried to resovle lower-bound environment, which was not necessary for the task.
It looks like it's because of the missing frozen flag...?

Failed test in the merge queue: https://github.com/scipp/ess/actions/runs/25638109487/job/75253633834

This failure is also weird that it sometimes passed when I re-ran the failed job...
I didn't want to look too much deep into this issue though.

## Pixi version
Default pixi version of the setup-pixi action version is not pinned per action version.
If this is not specified, it seems like it installs the latest version of pixi.
The setup-pixi action readme says it just picks up whatever the runner has...?

I didn't really understand what it does from the description though:
> You can also use a preinstalled local version of pixi on the runner by not setting any of the pixi-version, pixi-url or pixi-bin-path inputs. This action will then try to find a local version of pixi in the runner's PATH.

I think it'll be safer if we pin the version of the pixi in the CI first, and then we upgrade them once in a while...?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
           pixi-version: v0.67.2
           frozen: true
           environments: lint
-      - run: pixi run -e lint lint
+      - run: pixi run --frozen -e lint lint
 
   test:
     name: Test ${{ matrix.package }}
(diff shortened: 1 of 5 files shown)


YooSunYoung:
The git diff is messed up for some reason... and we're having more discussion about which pixi version to use/if the frozen flag actually works so... I'll open another PR from a u-messed up branch...

PR #561, opened 2026-05-11
Pin pixi version and set pixi environment variable in github action workflows.

YooSunYoung:
Continued from #559 

I tested how `pixi run` works after installation in #560.
And this was the best combination of the settings...

I still don't understand why the diff looks like this...??


.github/workflows/weekly-platform.yml
```diff
@@ -5,6 +5,10 @@ on:
   schedule:
     - cron: "0 2 * * 1"
 
+env:
+  PIXI_FROZEN: true
+  PIXI_NO_INSTALL: true
```
jl-wynen, line 10: What does NO_INSTALL do? Don't we need to install dependencies?

YooSunYoung: It only affects `pixi run`, not `pixi install`, which is done by `setup-pixi` action.

YooSunYoung:
Should we already use this?

https://pixi.prefix.dev/latest/reference/pixi_manifest/#exclude-newer-optional

jl-wynen:
> Should we already use this?
> 
> https://pixi.prefix.dev/latest/reference/pixi_manifest/#exclude-newer-optional

We should. But we don't auto-update dependencies. So can we do more than recommend people to use it?

jl-wynen:
@MridulS Do you have any inputs on this?

MridulS (approved):

MridulS:
> https://pixi.prefix.dev/latest/reference/pixi_manifest/#exclude-newer-optional

~I think we need to wait a bit as there is no way of excluding packages (like scipp, plopp) from the exclude-newer and we won't be able to get scipp releases as a deps quickly.~

Ignore, they fixed it!

PR #582, opened 2026-05-11
[CI] Exclude newer in pixi toml.

### LINKED ISSUE #271: Cool down period for updating dependencies.
Attempted at #266 

I think it's still a good idea so we should add this policy somehow (or just wait until pixi supports it...)

YooSunYoung:
Fixes #271

I randomly chose `14d` but we can adjust it...


.github/workflows/pr-vibe-check.yml
YooSunYoung, line 1: This change is not related to the PR but it was easier to fix here so that we also test it...

pixi.toml
```diff
@@ -11,6 +12,16 @@ python = "3.11.*"
...
+[exclude-newer]
+scipp = "0m"
+scippneutron = "0m"
+scippnexus = "0m"
+
+[pypi-exclude-newer]
+scipp = "0m"
+scippneutron = "0m"
+scippnexus = "0m"
+
```
MridulS, line 26: plopp too? or we don't have a hard dependency on plopp anywhere?

YooSunYoung: Plopp is hard dependency of almost all packages I think

MridulS (approved):

PR #590, opened 2026-05-18
[ESSDIFFRACTION] FEAT: Add scaffolding for MAGIC

MridulS:


packages/essdiffraction/src/ess/magic/types.py
```diff
@@ -0,0 +1,31 @@
...
+
+class PolarizationState(Enum):
+    up = 'up'
+    down = 'down'
+
+
+IncidentPolarization = NewType("IncidentPolarization", sc.Variable)
+"""Polarisation of the incident beam."""
+
+ScatteredPolarization = NewType("ScatteredPolarization", sc.Variable)
```
jokasimr, line 30: There might already be domain types in the `esspolarization` package that can be used here.

But this can also wait until we know more what is needed here.

jokasimr (approved):
👍

Issue #591, opened 2026-05-20
Nightly tests and artefact uploading


jl-wynen:
The nightly tests are still running in the old repos. And they are uploading reduced data files which are used by the DMSC nightly tests. We need to

- Disable nightlies in old repos
- Upload reduced files in `ess` nightlies (DREAM, ESTIA)
- Make sure those files are picked up by the DMSC nightly pipeline

nvaytet:
We chatted about this today.
We would like to get around the uploading of artifact and then using another service to run a gh command to find the latest artifact and publish it on a stable url.

We will try to instead use a gh pre-release to publish the results (tried this today on the old `essreflectometry` repo.

MridulS:
- [x] https://github.com/scipp/ess/pull/596 
- [x] Update Analysis team and DMSC nightly pipeline with the new URLs
       - https://git.esss.dk/dmsc-nightly/dmsc-nightly/-/merge_requests/203
       - https://github.com/easyscience/reflectometry-lib/pull/365
       - https://github.com/easyscience/diffraction-lib/pull/190
- [x] Disable old repo nightlies
      - Turned off actions on ess* (except esslivedata)
- [x] Delete VM and Cloudflare R2 bucket
      - Turned off the VM
      - Disabled Cloudflare R2

PR #596, opened 2026-05-21
[CI] Publish nightly reduced data files

MridulS:
First step for https://github.com/scipp/ess/issues/591


MridulS:
Seems to work fine https://github.com/scipp/ess/releases/tag/reduced_data_nightly

Integrated and analysis tests can fetch the files from stable urls like https://github.com/scipp/ess/releases/download/reduced_data_nightly/dream_reduced.cif going forward.

.github/workflows/nightly.yml
```diff
@@ -34,7 +44,48 @@ jobs:
...
+          if-no-files-found: warn
+
+  publish:
+    name: Publish nightly reduced data
+    needs: test
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
```
nvaytet: Do we even need to checkout here?

.github/workflows/nightly.yml
```diff
@@ -34,7 +44,48 @@ jobs:
...
+          shopt -s nullglob
+          files=(./artifacts/*)
+          if [ ${#files[@]} -eq 0 ]; then
+            echo "No reduced data files were produced; nothing to publish."
+            exit 0
+          fi
+          echo "Publishing ${#files[@]} file(s):"
+          printf '  %s\n' "${files[@]}"
+          gh release delete reduced_data_nightly --yes --cleanup-tag 2>/dev/null || true
+          gh release create reduced_data_nightly \
```
nvaytet: What happens if we skip the delete step and just try to push new files to an existing release?
For example, when uploading the built docs as assets to a release, we use a [different action](https://github.com/scipp/ess/blob/main/.github/workflows/release.yml#L120) that just add files to a release.
I wonder if you can replace files with that, and avoid having to create the release entirely (now that it has been created once)?

nvaytet: There seems to be an option for overwriting:
```overwrite      false      If an asset with the same name already exists, overwrite it.```

nvaytet: Another reason to maybe use the action to upload assets is that I seem to get a gh notification every time a release is made.

Just uploading would not add any noise to our inboxes.

MridulS:
This should work now: https://github.com/scipp/ess/releases/tag/reduced_data_nightly

We edit the release instead of a new release so shouldn't create any more notifications.

nvaytet (approved):

PR #598, opened 2026-05-21
[CI] fix label bot reporting already-present labels as newly added

### LINKED ISSUE #584: Incorrect comment from label bot
In #583, the label bot said that it added a label even though that label already existed. But it did add a title prefix and correctly said so.

MridulS:
Fixes #584

### DIFF
--- a/.github/workflows/pr-vibe-check.yml
+++ b/.github/workflows/pr-vibe-check.yml
@@ -28,6 +28,10 @@ jobs:
   check-pr:
     name: Check Labels
     runs-on: ubuntu-slim
+    permissions:
+      # Read access for `actions/checkout` and `gh pr view`.
+      contents: read
+      pull-requests: read
     outputs:
       all_labels: ${{ steps.all-labels.outputs.all_labels }}
       new_labels: ${{ steps.new-labels.outputs.new_labels }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-vibe-check.yml
```diff
@@ -56,12 +60,18 @@ jobs:
...
            all_labels: ${{ steps.all-labels.outputs.all_labels }}
+           GH_TOKEN: ${{ github.token }}
         run: |
-          EXISTING_LABELS='${{ toJson(github.event.pull_request.labels.*.name) }}';
+          # Query the labels currently on the PR rather than reading them from
+          # the event payload. The payload is a snapshot taken when the event
+          # was created, so labels the author adds right after opening the PR
+          # (a common case) would be missing from it and wrongly reported as
+          # newly added by the bot. See issue #584.
+          EXISTING_LABELS="$(gh pr view ${PR_NUM} --json labels --jq '.labels[].name')";
```
MridulS, line 70: I'm not a 100% sure if this change is actually a good idea. If the tag is added at the moment the PR is created it should be in the payload already so we don't need to use the gh cli.

YooSunYoung: We'll still have the same problem anyway if human is faster than this line to be executed, no?

YooSunYoung: Or we should just remove the commenting part. It's cute but makes things unnecessarily complicated I think....

MridulS: Yeah maybe just remove the autocommenting.

YooSunYoung:
I removed the PR bot commenting  in #613 . We can close this PR I guess?

PR #600, opened 2026-05-22
[ESSNMX] Save TOF 1d histogram

### LINKED ISSUE #267: [essnmx] Print/log/dump 1-D histogram along the time dimension.
From the discussion in #264

YooSunYoung:
Fixes #267

Command to test:
```
pixi run --frozen essnmx-reduce \
    --output-dir aux \
    --input-file ${HOME}/.cache/ess/nmx/small_nmx_nexus.hdf.zip.unzip/small_nmx_nexus.hdf \
    --output-file test.hdf \
    --overwrite --nbins 50
```

Then it should save png file in the `aux/` directory and `test.hdf` should have `/entry/aux/tof-1d`.

### DIFF
--- a/packages/essnmx/src/ess/nmx/executables.py
+++ b/packages/essnmx/src/ess/nmx/executables.py
@@ -396,7 +396,6 @@ def save_auxiliary_output(
     # Export TOF 1D distribution into entry/aux
     if (
         output_config
-        and isinstance(tof_histogram, sc.DataArray)
         and not output_config.skip_file_output
         and not aux_config.no_tof_1d_in_file
     ):
--- a/packages/essnmx/src/ess/nmx/executables.py
+++ b/packages/essnmx/src/ess/nmx/executables.py
@@ -238,18 +238,11 @@ def reduction(
     # Check the file output configuration before we start heavy computation.
     if not config.output.skip_file_output:
         _check_file(config.output.output_file, config.output.overwrite)
-    elif not (config.output.skip_file_output or config.aux.no_axilaries):
         config.aux.check_output_dir()
 
     display = _retrieve_display(logger, display)
     input_file_path = _retrieve_input_file(config.inputs.input_file).resolve()
     display(f"Input file: {input_file_path}")
-
-    output_file_path = pathlib.Path(config.output.output_file).resolve()
-    display(f"Output file: {output_file_path}")
-
-    auxilary_output_dir = config.aux.build_target_dir(output_file_path.as_posix())
-    display(f"Auxiliary output dir: {auxilary_output_dir}")
 
     detector_names = select_detector_names(detector_ids=config.inputs.detector_ids)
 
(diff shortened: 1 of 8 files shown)


YooSunYoung:
~Hmm I think I'll change where the auxiliary stuff is saved in the nxlauetof file....~ Done...!

packages/essnmx/tests/executable_output_test.py
```diff
@@ -107,3 +112,66 @@ def test_compare_output_file_with_frozen(tmp_path: pathlib.Path):
...
+            output_file=output_file.as_posix(),
+            compression=Compression.NONE,
+            skip_file_output=False,
+        ),
+        aux=AuxiliaryOutputConfig(no_png=False, no_tof_1d_in_file=False),
+    )
+    with pytest.warns(RuntimeWarning, match="No crystal rotation*"):
+        results = reduction(config=config)
+
+    # Test PNG file.
```
YooSunYoung:
```suggestion
    # Test if the PNG file is generated.
```

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -357,6 +371,106 @@ def save_results(*, results: NMXLauetof, output_config: OutputConfig) -> None:
...
+    tof_histogram = sc.reduce(
+        det_res.data.sum(["x_pixel_offset", "y_pixel_offset"])
+        for det_res in results.instrument.detectors.values()
+        if isinstance(det_res.data, sc.DataArray)
+    ).sum()
+
+    # Export TOF 1D distribution into entry/aux
+    if (
+        output_config
+        and isinstance(tof_histogram, sc.DataArray)
```
jokasimr: What else might it be here, might it be `None`?

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -357,6 +371,106 @@ def save_results(*, results: NMXLauetof, output_config: OutputConfig) -> None:
...
+        and not output_config.skip_file_output
+        and not aux_config.no_tof_1d_in_file
```
jokasimr: Why is there an option to disable it? Is it costly to produce the figure?

jokasimr (commented):
My only concern is that the more options the interface has the more complex will it be to work with and the more cumbersome it will be to modify (because removing /changing the options might break some user).

For example, we could just say that we always export the png and write the tof curve to the file. It doesn't seem that important to be able to disable those.

Other than that it looks good to me. 👍

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: No... it can't be anything else... I'll remove it.

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: Not really...
It'll be at maximum thousands of floating numbers.

But it's always good to be able to turn off the auxiliary related steps I thought.
Also makes it more clear for users that there is an auxiliary information in the output from the configuration itself.

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: Done..!

packages/essnmx/src/ess/nmx/executables.py
jokasimr: Not sure I agree it's always good to have options. I'd lean towards fewer options until they're explicitly requested 🤔 but I'll leave it to you to decide.

jokasimr (approved):

packages/essnmx/src/ess/nmx/executables.py
YooSunYoung: Yeah... I was a bit lazy to update the regression test and some other related tests but I did update it properly so that auxiliary output doesn't break those tests,
and then removed the `no_*` configuration options.
We can add the options back when we want.
Thanks!

Issue #599, opened 2026-05-22
Run doctest and linkcheck again


jl-wynen:
The new pixi setup only builds the docs but does not test them like we used to. We should run doctest and linkcheck in our CI.

jl-wynen:
#601 implements the doctest part of this. We used to run linkcheck only on main because it can be flaky and we didn't want to block PRs. But we no longer run the pipeline on main. How can we best do this?

PR #601, opened 2026-05-22
[CI] Unified docs CI workflow

### LINKED ISSUE #265: Make a reusable docs CI workflow
We have 3 workflows that build docs, 2 of them also publish the docs to different places. Can we merge these into a reusable workflow and call that from CI, releases, and the web UI?

_Originally posted by @jl-wynen in https://github.com/scipp/ess/pull/262#discussion_r2981729524_

jl-wynen:
This combines the different CI workflows into a single, reusable docs workflow.

Fixes #265
Part of #599 (only doctest)


.github/workflows/release.yml
YooSunYoung: Should this change to `docs-${{ inputs.package }}`?

jl-wynen: Good catch!

YooSunYoung (approved):

PR #603, opened 2026-05-22
[essreduce] On-the-fly wavelength lookup table in generic workflow

nvaytet:
Insert the providers from the LUT workflow into the GenericUnwrapWorkflow to compute the lut on the fly with chopper cascade.
Breaking change: we compute a LUT per component (detector bank, monitor) instead of one table for all.

Building small tables is faster, and interpolating inside a small table is also faster (most probably due to cache misses in larger tables).

By default, the workflow tries to load the lookup table from file, which behaves like the old code.
The workflow now has 3 modes it can operate in:

- `wavelength_from='file'`: the default = read pre-computed table from file
- `wavelength_from='simulation'`: on-the-fly computed table using tof simulation from chopper params (slow)
- `wavelength_from='analytical'`:  on-the-fly computed table using chopper_cascade -> what we want to default to in the future.

Suggestions for a better name than `'analytical'` are welcome.

The idea is that each package would then move over to `analytical` mode when they wish to do so (in follow-up PRs).

<hr>

**Additional notes:**

I tried to reduce the number of parameters we need to set on the workflow:

- `LtotalRange` is now determined from the pixel positions of the detector bank, or the monitor position.
- `PulseStride` is now guessed from the chopper frequencies (see [comment here](https://github.com/scipp/ess/pull/589#discussion_r3259448231))

Finally, the `LookupTableWorkflow` that was used to create lookup tables using `tof` is now just an alias for the `GenericUnwrapWorkflow` that uses `tof` by default, so that the behaviour remains the same as before. We plan to remove the alias in the future.


packages/essnmx/tests/executable_test.py
```diff
@@ -561,42 +564,32 @@ def test_reduction_compression_gzip(
...
 
-    # GZIP is expected to have better compression ratio than BITSHUFFLE
     assert (
-        file_paths[Compression.BITSHUFFLE_LZ4].stat().st_size
-        > file_paths[Compression.GZIP].stat().st_size
+        file_paths[Compression.NONE].stat().st_size
+        > file_paths[Compression.BITSHUFFLE_LZ4].stat().st_size
     )
-    # BITSHUFFLE is expected to be faster than GZIP
-    assert total_times[Compression.BITSHUFFLE_LZ4] < total_times[Compression.GZIP]
```
nvaytet, line 593: This makes the test flaky, as sometimes other factors than just the compression can influence timings of the entire reduction.
In addition, I don't think we should be testing speed of compression algorithms in our unit tests, that's more the job of the makers of the compression algorithms. Testing that we can write compressed data and that it can be read and decompressed again is enough.

packages/essnmx/src/ess/nmx/executables.py
```diff
@@ -318,9 +325,6 @@ def reduction(
         sample=sample_meta,
     )
 
-    if config.workflow.time_bin_coordinate == TimeBinCoordinate.time_of_flight:
-        results.lookup_table = base_wf.compute(LookupTable)
```
nvaytet, line 322: For now, it is unclear if we want to store a table for each detector panel as an attribute on the `results` object. It was only the in-memory object, as the LUT is not saved to disk alongside the reduced data.

It was only used for inspection, but no code depended on it. So we just removed it.

packages/essnmx/src/ess/nmx/workflows.py
```diff
@@ -37,68 +36,15 @@
...
     NMXSourceMetadata,
     TofDetector,
-    TofSimulationMaxWavelength,
-    TofSimulationMinWavelength,
 )
 
 default_parameters = {
-    TofSimulationMaxWavelength: sc.scalar(3.6, unit='angstrom'),
-    TofSimulationMinWavelength: sc.scalar(1.8, unit='angstrom'),
+    SimulationMaxWavelength: sc.scalar(3.6, unit='angstrom'),
```
nvaytet, line 42: I removed the `Tof` qualifier, as this can also apply to the analytical 'chopper cascade' simulation.

packages/essnmx/src/ess/nmx/workflows.py
```diff
@@ -37,68 +36,15 @@
...
+    SimulationMinWavelength: sc.scalar(1.8, unit='angstrom'),
     LookupTableRelativeErrorThreshold: {f'detector_panel_{i}': 0.1 for i in range(5)},
 }
 
 
-def _simulate_fixed_wavelength_tof(
```
nvaytet, line 51: This function was replaced by just using the generic workflow with `wavelength_from='simulation'`.
We also no longer need to merge workflows below.

packages/essnmx/src/ess/nmx/workflows.py
```diff
@@ -37,68 +36,15 @@
...
-                wavelength=events.coords["wavelength"],
-                weight=events.data,
-                distance=source.distance,
-            )
-        },
-        choppers=None,
-    )
-
-
-def _merge_panels(*da: sc.DataArray) -> sc.DataArray:
```
nvaytet, line 96: It seems this function was not used?

packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -52,7 +62,7 @@ def __post_init__(self):
 
 
 @dataclass
-class SimulationResults:
+class SimulationResultsBaseClass:
```
nvaytet: Any suggestions for a better name, or an alternative way to avoid needing both `SimulationResultsBaseClass` (the dataclass) and `SimulationResults` (the `RunType` Scope) are welcome!

packages/essreduce/src/ess/reduce/unwrap/to_wavelength.py
```diff
@@ -448,73 +450,26 @@ def _mask_large_uncertainty_in_lut(
...
-    mask = relative_error > sc.scalar(error_threshold)
-    return LookupTable(
-        **{
-            **asdict(table),
-            "array": sc.where(mask, sc.scalar(np.nan, unit=da.unit), da),
-        }
-    )
-
-
-def mask_large_uncertainty_in_lut_detector(
```
nvaytet, line 463: This and the 'monitor' version below were simply replaced by just using the generic `mask_large_uncertainty_in_lut` that operates on `RunType`. The type specialization happens automatically further down the pipeline.

nvaytet:
@YooSunYoung can you double-check I didn't mess anything up in `essnmx`? :crossed_fingers:

packages/essreduce/docs/user-guide/unwrap/dream.ipynb
```diff
@@ -341,14 +347,14 @@
...
-    "lut_wf[DiskChoppers[AnyRun]] = disk_choppers\n",
-    "lut_wf[SourcePosition] = source_position\n",
-    "lut_wf[LtotalRange] = (\n",
+    "lut_wf = LookupTableWorkflow(run_types=[SampleRun, BackgroundRun])\n",
+    "lut_wf[DiskChoppers[SampleRun]] = disk_choppers\n",
+    "lut_wf[Position[snx.NXsource, SampleRun]] = source_position\n",
+    "lut_wf[LtotalRange[SampleRun, snx.NXdetector]] = (\n",
     "    sc.scalar(25.0, unit=\"m\"),\n",
     "    sc.scalar(80.0, unit=\"m\"),\n",
     ")\n",
```
SimonHeybrock, line 356: I think we chatted about this, but remind me why this needs to be set. It seems the range is not per-component, so we is it not computed?

packages/essreduce/src/ess/reduce/unwrap/lut.py
SimonHeybrock: Why do you need both instead of making the dataclass generic?

packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -82,18 +116,24 @@ class SimulationResults:
...
+    Range (min, max) of the total length of the flight path from the source to the
+    detector. This is used to create the lookup table to compute the neutron
+    time-of-flight. Note that the resulting table will extend slightly beyond this
+    range, as the supplied range is not necessarily a multiple of the distance
+    resolution.
+
+    Note also that the range of total flight paths is supplied manually to the workflow
+    instead of being read from the input data, as it allows us to compute the expensive
+    part of the workflow in advance (the lookup table) and does not need to be repeated
+    for each run, or for new data coming in in the case of live data collection.
```
SimonHeybrock: I think this answers my comment above... partially. But live data collection caches static nodes (using `StreamProcessor`), so this is not a reason, I think?

packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -733,61 +798,153 @@ def make_wavelength_lut_from_polygons(
...
+    return LookupTable[RunType, Component](
+        Lut(
+            array=table,
+            pulse_period=pulse_period,
+            pulse_stride=pulse_stride,
+            distance_resolution=table.coords["distance"][1]
+            - table.coords["distance"][0],
+            time_resolution=table.coords["event_time_offset"][1]
+            - table.coords["event_time_offset"][0],
+            # TODO: Do we still want to store the chopper info in the lookup table?
```
SimonHeybrock: Planned to resolve here or later?

packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -733,61 +798,153 @@ def make_wavelength_lut_from_polygons(
...
+    if wavelength_from == "file":
+        return params
+
+    params.update(
+        {
+            PulsePeriod: 1.0 / sc.scalar(14.0, unit="Hz"),
+            DistanceResolution: sc.scalar(0.1, unit="m"),
+            TimeResolution: sc.scalar(250.0, unit='us'),
+            SourceBounds: SourceBounds(
+                time=(sc.scalar(0.0, unit='ms'), sc.scalar(5.0, unit='ms')),
```
SimonHeybrock, line 910: Where does 5.0 come from?

packages/essnmx/tests/executable_test.py
```diff
@@ -561,42 +564,32 @@ def test_reduction_compression_gzip(
...
+def test_reduction_compression_bitshuffle(
     reduction_config: ReductionConfig, tmp_path: pathlib.Path
 ) -> None:
     reduction_config.output.skip_file_output = False
     reduction_config.workflow.nbins = 5  # For faster test
     file_paths: dict[Compression, pathlib.Path] = {}
-    total_times: dict[Compression, pathlib.Path] = {}
 
-    for compress_mode in (Compression.GZIP, Compression.BITSHUFFLE_LZ4):
+    for compress_mode in (Compression.NONE, Compression.BITSHUFFLE_LZ4):
```
YooSunYoung, line 574: Why did this one removed?

packages/essnmx/tests/executable_test.py
YooSunYoung: It was to be clear about what we expect from different compression modes.
Also, we wanted to see how consistently fast the bitshuffle algorithm is...
I agree the pytest framework makes the time consumption test flaky.

We can remove the timing assertion but we shouldn't remove the GZIP compression mode from this test.

packages/essdiffraction/src/ess/dream/workflows.py
```diff
@@ -97,7 +98,9 @@ def _collect_reducer_software() -> ReducerSoftware:
     )
 
 
-def DreamWorkflow(**kwargs) -> sciline.Pipeline:
+def DreamWorkflow(
+    wavelength_from: Literal["analytical", "simulation", "file"] = "file", **kwargs
```
jl-wynen: Given that the `wavelength_from` arg needs to be repeated many times, can you define a type in the unwrap package? Either a type alias for the `Literal` or a `StrEnum`.

packages/essdiffraction/src/ess/dream/workflows.py
```diff
@@ -109,19 +112,27 @@ def DreamWorkflow(**kwargs) -> sciline.Pipeline:
 
     Parameters
     ----------
+    wavelength_from:
+        Mode for creating the wavelength lookup table. The 'analytical' mode uses
+        analytical calculations to propagate and chop a pulse through the chopper
+        cascade and build the lookup table. The 'simulation' mode uses ``tof`` to trace
+        individual neutrons through the chopper system and build the table.
+        The 'file' mode loads a pre-computed table from a file.
```
jl-wynen: Instead of explaining this here, can you link to a docs page that explains the 3 modes in detail?

packages/essdiffraction/tests/dream/geant4_reduction_test.py
```diff
@@ -122,24 +123,24 @@ def dream_lookup_table():
...
         sc.scalar(60.0, unit="m"),
         sc.scalar(80.0, unit="m"),
     )
     lut_wf[unwrap.DistanceResolution] = sc.scalar(0.1, unit="m")
     lut_wf[unwrap.TimeResolution] = sc.scalar(250.0, unit='us')
-    return lut_wf.compute(unwrap.LookupTable)
+    return lut_wf.compute(unwrap.LookupTable[AnyRun, NXdetector])
 
 
 def test_pipeline_can_compute_dspacing_result_using_custom_built_tof_lookup(
```
jl-wynen: Now we should also have tests with a table that is computed on the fly. Is it easy to parametrize a test?

packages/essimaging/tests/odin/data_reduction_test.py
```diff
@@ -27,9 +28,14 @@ def workflow() -> sl.Pipeline:
...
-    # Cache the lookup table
-    wf[LookupTable] = wf.compute(LookupTable)
+    for run_type in (SampleRun, OpenBeamRun):
+        wf[LookupTableFilename[run_type, NXdetector]] = (
+            odin.data.odin_wavelength_lookup_table()
+        )
+        # Cache the lookup table
+        wf[LookupTable[run_type, NXdetector]] = wf.compute(
+            LookupTable[run_type, NXdetector]
+        )
```
jl-wynen, line 38: Is this possible? I.e., can we set a lut for all runs without looping through them?

```suggestion
    wf[LookupTableFilename[RunType, NXdetector]] = (
        odin.data.odin_wavelength_lookup_table()
    )
    # Cache the lookup table
    wf[LookupTable[RunType, NXdetector]] = wf.compute(
        LookupTable[RunType, NXdetector]
    )
```

packages/essreduce/docs/user-guide/unwrap/frame-unwrapping.ipynb
```diff
@@ -209,9 +210,13 @@
...
    "source": [
-    "lut_wf = LookupTableWorkflow()\n",
-    "lut_wf[LtotalRange] = detectors[0].distance, detectors[-1].distance\n",
-    "lut_wf[DiskChoppers[AnyRun]] = {\n",
+    "lut_wf = LookupTableWorkflow(run_types=[SampleRun])\n",
+    "lut_wf[Position[snx.NXsource, SampleRun]] = sc.vector([0, 0, 0], unit=\"m\")\n",
+    "lut_wf[LtotalRange[SampleRun, snx.NXdetector]] = (\n",
+    "    detectors[0].distance,\n",
+    "    detectors[-1].distance,\n",
+    ")\n",
```
jl-wynen, line 218: Why do we need both a source position and Ltotal range? And why is the source in the origin? Is this about determining the chopper distances?

packages/essreduce/src/ess/reduce/unwrap/types.py
```diff
@@ -10,14 +10,16 @@
...
-LookupTableFilename = NewType("LookupTableFilename", str)
-"""Filename of the wavelength lookup table."""
+
+class LookupTableFilename(sl.Scope[RunType, Component, str], str):
+    """Filename of the wavelength lookup table."""
 
 
 @dataclass
-class LookupTable:
+class Lut:
```
jl-wynen: I find it confusing to have both `Lut` and `LookupTable`. Is there a better name for the dataclass?

Or can you make `LookupTable` the dataclass with
```python
class ErrorLimitedLookupTable(sl.Scope[RunType, Component, LookupTable], LookupTable):
```
?

packages/essreduce/src/ess/reduce/unwrap/lut.py
jl-wynen: But you still need to repeat this for every `RunType` which usually have the same detector positions. Or does the workflow still need to do that anyway?

packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -223,14 +270,14 @@ def _compute_mean_wavelength(
     return mean_wavelength
 
 
-def make_wavelength_lookup_table(
-    simulation: SimulationResults,
-    ltotal_range: LtotalRange,
+def make_wavelength_lut_from_simulation(
```
jl-wynen, line 244: How about 'monte_carlo' instead of 'simulation' to be more precise? Not sure what to do about 'analytical', though.

packages/essreduce/src/ess/reduce/unwrap/lut.py
nvaytet: Later

packages/essreduce/docs/user-guide/unwrap/frame-unwrapping.ipynb
nvaytet: > Is this about determining the chopper distances?

Yes, for chopper distances we need a source position.

packages/essreduce/src/ess/reduce/unwrap/lut.py
SimonHeybrock: We cannot have two runs at the same time in live data. If we need to subtract vanadium etc. then that must come from a previously stored intermediate result.

packages/essreduce/docs/user-guide/unwrap/frame-unwrapping.ipynb
jl-wynen: And why is it 0? It looked like in other workflows, it comes from the input file and then the sample is in the origin.

SimonHeybrock:
Found a crash in the new `analytical` mode while integrating it into ESSlivedata: `make_wavelength_lut_from_polygons` raises `IndexError` on a degenerate subframe when a chopper has a **positive** `frequency`.

```
File ".../ess/reduce/unwrap/lut.py", line 554, in _polygon_intersections
    if b[0, 0] == b[1, 0]:
                  ~^^^^^^
IndexError: index 1 is out of bounds for axis 0 with size 1
```

A polygon bound decomposes to a single vertex (the subframe collapses), so indexing `b[1]` is out of range.

### Minimal reproducer (essreduce only)

```python
import scipp as sc
import scippnexus as snx
from scippneutron.chopper import DiskChopper
from ess.reduce import unwrap
from ess.reduce.unwrap import GenericUnwrapWorkflow
from ess.reduce.nexus.types import AnyRun, Position


def chopper(z, freq):
    return DiskChopper(
        axle_position=sc.vector([0, 0, z], unit='m'),
        frequency=sc.scalar(freq, unit='Hz'),
        beam_position=sc.scalar(0.0, unit='deg'),
        phase=sc.scalar(0.0, unit='rad'),
        slit_begin=sc.array(dims=['cutout'], values=[0.0], unit='deg'),
        slit_end=sc.array(dims=['cutout'], values=[90.0], unit='deg'),
        slit_height=None,
        radius=sc.scalar(0.35, unit='m'),
    )


def run(freq):
    wf = GenericUnwrapWorkflow(
        run_types=[AnyRun], monitor_types=[], wavelength_from='analytical'
    )
    wf[unwrap.DiskChoppers[AnyRun]] = sc.DataGroup(
        {'chopper1': chopper(-15.0, freq), 'chopper2': chopper(-13.0, freq)}
    )
    wf[Position[snx.NXsource, AnyRun]] = sc.vector([0, 0, -25.0], unit='m')
    wf[unwrap.PulseStride[AnyRun]] = 1
    wf[unwrap.LtotalRange[AnyRun, snx.NXdetector]] = (
        sc.scalar(5.0, unit='m'),
        sc.scalar(30.0, unit='m'),
    )
    wf[unwrap.DistanceResolution] = sc.scalar(0.1, unit='m')
    return wf.compute(unwrap.LookupTable[AnyRun, snx.NXdetector])


run(-14.0)  # OK
run(+14.0)  # IndexError
```

### Trigger

The sign of `frequency` is what flips it: `-14 Hz` builds the table fine, `+14 Hz` crashes. Everything else identical. The `simulation` mode tolerated this geometry (it just yields sparse neutrons), so it only surfaced on the analytical path.

Real beamline choppers use the signed (negative, anti-clockwise) convention, so this does not bite in practice — our LOKI geometry computes fine. We hit it only with a synthetic test fixture that fed `+14 Hz`. Reporting because a positive frequency arguably shouldn't produce an unguarded `IndexError`; a degenerate-subframe guard or an explicit error would be friendlier. Happy to leave the fix to you — on our side we'll just fix the fixture's sign.

packages/essnmx/tests/executable_test.py
nvaytet: > we shouldn't remove the GZIP compression mode from this test

Not sure I entirely follow what you wanted to test in this test.
The Gzip mode is tested in the test just above.

I think testing who of Gzip or bitshuffle produces the smallest results shouldn't be part of testing that the executable runs correctly? I think we should just test that both can be used, and that the data can be read afterwards?
I don't think it's worth verifying the compression performance, that is a promise made by the libraries implementing the compression and not something we need to compare in our unit tests?

packages/essreduce/src/ess/reduce/unwrap/lut.py
nvaytet: Can you show me how you would do that?
I thought I needed to first make a Dataclass and then use that in a Scope...

packages/essreduce/docs/user-guide/unwrap/dream.ipynb
nvaytet: It no longer needs to be set.

It is however useful to keep as an entry point in the workflow in case we wish to compute a lookup table that covers a wide range of distances with an expensive calculation using tof or McStas.

In the present example, instead I will set the `DetectorLtotal` (which is defined further up in the notebooks).

packages/essreduce/src/ess/reduce/unwrap/lut.py
nvaytet: This comment is actually a remnant of when we were only using tables from disk or computed using `tof`.
I think I will just remove the comment.

packages/essreduce/src/ess/reduce/unwrap/lut.py
nvaytet: I added a comment.

packages/essreduce/src/ess/reduce/unwrap/lut.py
SimonHeybrock: `sciline.Scope` is just a hack for when you want to define an alias for a common type such as `int` or `scipp.DataArray`. If you anyway have a special type (such as a `dataclass` you can just use `Generic` directly, not need for `Scope`.

packages/essreduce/docs/user-guide/unwrap/dream.ipynb
nvaytet: Note that my plan is to probably remove the `dream.ipynb` and `wfm.ipynb` notebooks here; the new `analytical-unwrap.ipynb` is meant to replace them.

packages/essdiffraction/src/ess/dream/workflows.py
nvaytet: I went with `StrEnum`.

packages/essdiffraction/tests/dream/geant4_reduction_test.py
nvaytet: Yes, I parametrized most tests for `essreduce.unwrap`.
Do you want me to parametrize all such tests in the child packages?

packages/essimaging/tests/odin/data_reduction_test.py
nvaytet: Yes, this is the feature of `sciline` I stumbled upon this morning that I didn't know existed.

nvaytet: Hmm doesn't seem to work, I get `UnsatisfiedRequirement` when I try to compute `LookupTable[RunType, NXdetector]`

packages/essreduce/docs/user-guide/unwrap/frame-unwrapping.ipynb
nvaytet: I don't know, it's just a beamline where I make everything up (choppers, detector positions).
It's always been 0, unrelated to these changes.

packages/essreduce/src/ess/reduce/unwrap/types.py
nvaytet: I think it's related to @SimonHeybrock 's comment above: I need a `LookupTable` also per `RunType` and `Component`. So I thought I needed to make a Dataclass and then make `LookupTable` as the Scope.

But maybe you can show me a better way it can be achieved?

nvaytet:
I added
```Py
if len(b) < 2:
    continue
```
inside the loop. Does that fix it for you?

packages/essreduce/src/ess/reduce/unwrap/lut.py
nvaytet: ~I am not sure I can use the same thing for `Lut`, `LookupTable`, and `ErrorLimitedLookupTable`? I feel I would have to repeat the implementation of the dataclass for both (unless I missed something?)~

Scratch that, I think I found how to do it by just defining
```Py
class ErrorLimitedLookupTable(LookupTable[RunType, Component]):
```

packages/essdiffraction/src/ess/dream/workflows.py
nvaytet: Ok I will do that.

packages/essreduce/src/ess/reduce/unwrap/types.py
nvaytet: I fixed it, see above.

packages/essdiffraction/tests/dream/geant4_reduction_test.py
jl-wynen: There is some custom code to handle lut filenames, so ideally, they should be parametrized. But that would lead to additional reference data because the workflows don't produce exactly the same result for each lut, right?

packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -21,9 +21,10 @@
     DetectorLtotal,
     LookupTable,
     LookupTableFilename,
-    Lut,
+    # Lut,
```
jl-wynen: Remove?

packages/essimaging/tests/odin/data_reduction_test.py
jl-wynen: Right, that can't work. Maybe 
```python
    wf[LookupTable[RunType, NXdetector]] = wf.compute(
        LookupTable[SampleRun, NXdetector]
    )
```
if you are sure the tables are all the same.

But this is only in a test. I was after a general approach to simplify code. In real workflows
```python
    wf[LookupTableFilename[RunType, NXdetector]] = (
        odin.data.odin_wavelength_lookup_table()
    )
```
should be all you need. And that should work.

packages/essreduce/src/ess/reduce/unwrap/types.py
jl-wynen: I like your new solution :+1:

packages/essdiffraction/src/ess/dream/workflows.py
nvaytet: See `lut-building-methods.ipynb`.

packages/essimaging/tests/odin/data_reduction_test.py
nvaytet: Maybe even
```Py
wf[LookupTableFilename] = odin.data.odin_wavelength_lookup_table()
```
?

packages/essdiffraction/src/ess/dream/workflows.py
nvaytet: https://remote-unzip.scipp.deno.net/scipp/ess/artifacts/7364081348/user-guide/unwrap/lut-building-methods.html

packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -547,6 +547,10 @@ def _polygon_intersections(polygons: list[np.ndarray], x: np.ndarray) -> np.ndar
         # To fix, if the two leftmost or rightmost points have the same x value, we set
         # the y value of the first point to be the same as the second point.
         for b in (bound1, bound2):
+            if len(b) < 2:
+                # This can happen if the polygon is degenerate (collapsed to a single
+                # vertex).
+                continue
```
SimonHeybrock: Are you implying that the code below that also uses `b` and the bounds works nevertheless?

nvaytet: I think it does?
But maybe you think it's safer to just discard the polygon if it has collapsed?

Can you clarify what a collapsed polygon means? Should there just be no polygon at all, or is there a single infinitessimally small point where neutrons could get through? (maybe it amounts to the same anyway)

SimonHeybrock: Good question. Does `scippneutron` keep around such degenerate polygons? Should they be dropped?

packages/essdiffraction/tests/dream/geant4_reduction_test.py
nvaytet: I parametrized the dream geant4 tests.
I suggest we update other tests once packages individually move over to using the `analytical` mode?

SimonHeybrock (approved):

packages/essreduce/src/ess/reduce/unwrap/lut.py
nvaytet: From private conversation:

So I looked at the subframes in your exampl (Minimal reproducer).
the subframe that is still there is not just a single vertex, it's made up of 3 vertices and they are all the same: [0, 0, 0] in both time and wavelength.

It's a peculiarity that happens because we allow a 0 wavelength (the default bounds of the pulse) which is infinite speed.
The choppers are open just before t=0 and close exactly at t=0.
I'm thinking that the neutrons with infinite speed can make it through just at t=0, but neutrons with infinite speed that started at t=1ms don't make it through because they were born too late.
So we are left with a single point at the corner of the pulse that can make it through.

If I change the pulse bounds to start at 0.001 angstrom the error goes away.
I get a new error which crashes when the subframes are empty, but I also fixed that now.

jl-wynen (approved):

packages/essnmx/tests/executable_test.py
YooSunYoung: @nvaytet The test was about to check if the configuration gives the correct, expected result at the end as it promised in the user-guide.
I disagree that we shouldn't verify the compression results. We can't just blindly trust what the upstream library claims that they are or the current benchmark results with specific version of the dependencies and promise the expected output to the users in our documentation.

And if verifying the compression performance really shouldn't be part of our tests, we might as well just remove all these comparison parts. There's no need to compare it to the None compression results either in that case.

nvaytet: Any opinions @SimonHeybrock @jl-wynen ? as we don't seem to agree here...

I'm worried that a third-party library version update could suddenly start to make the test fail even though we have not changed anything. Or if it starts behaving differently on a different file system or architecture if someone runs the tests locally on a new machine or new version of gh runners get deployed... Meaning we would have to spend time tracking down why it is now failing.

I already had to do so for the timings in the test, I was just trying to improve the stability of the tests. I didn't think this was such a sensitive issue.

> as it promised in the user-guide.

We don't have to make any performance promises, we can refer with links to the different algorithms what the promises are?

> There's no need to compare it to the None compression results either in that case.

I think checking that the compression is at least doing something compared to no compression is safe enough to verify?

PR #604, opened 2026-05-27
[CI] ci: fix docs deploy

jokasimr:


.github/workflows/docs.yml
```diff
@@ -79,32 +79,39 @@ jobs:
...
     needs: [ docs ]
+    if: ${{ inputs.publish }}
     runs-on: 'ubuntu-24.04'
     permissions:
       contents: write
     steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: ${{ inputs.branch == '' && github.ref || inputs.branch }}
+          fetch-depth: 0
```
YooSunYoung:
```suggestion
```
Do we need fetch-depth to be 0...? It takes more than a minute to fetch all history.

YooSunYoung (approved):

PR #605, opened 2026-05-28
Move dask to optional/test dependencies

nvaytet:
Move Dask to be an optional dependency.

There is a naive scheduler in Sciline which doesn't need Dask, and Dask pulls in many dependencies which is not always desirable.

That said, the Dask scheduler is smarter and manages memory better, so Dask should probably be installed on any reduction environment anyway.

I also standardized indentation in the pyproject files.


nvaytet:
In the end, after discussion, it was decided to keep Dask as a dependency as all other `ess*` packages depend on it.
The Dask scheduler is expected to give better performance in most cases.

PR #606, opened 2026-05-28
ci: add lower bound and latest released nightlies

jokasimr:
Adds two nightly test runs:

1. "Latest dependencies" corresponds to the old "Nightly tests at main branch"
2. "Lower bound" corresponds to the old "Nightly tests using lower bound dependencies"


.github/workflows/nightly.yml
```diff
@@ -54,10 +54,40 @@ jobs:
...
-      - uses: prefix-dev/setup-pixi@v0.9.4
+      - uses: actions/setup-python@v6
         with:
-          pixi-version: v0.68.0
-          frozen: true
-          environments: lb-${{ matrix.package }}
+          python-version: "3.11"
+      - uses: astral-sh/setup-uv@v8
+        with:
+          version: "0.7.21"
```
YooSunYoung: It looks like `astral-sh/setup-uv@v8` doesn't work...?
https://github.com/scipp/ess/actions/runs/26565652050/job/78259509158

Also it's recommended to use the commit ID for the github action so we should probably do this:

```suggestion
      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
        with:
          version: "0.7.21"
          python-version: "3.11"
```

I also merged the python setup into setup-uv in the suggestion.

.github/workflows/nightly.yml
```diff
@@ -54,10 +54,40 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+      - uses: astral-sh/setup-uv@v8
+        with:
+          version: "0.7.21"
```
YooSunYoung:
```suggestion
      - uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
        with:
          version: "0.7.21"
          python-version: "3.11"
```
Here as well.

YooSunYoung (changes requested):
I tried triggering it but it failed with this error message:
```
Error: Unable to resolve action `astral-sh/setup-uv@v8`, unable to find version `v8`
```
https://github.com/scipp/ess/actions/runs/26565652050/job/78259509158

I found that the major/minor tags anymore.
The explanation was in the release note.
https://github.com/astral-sh/setup-uv/releases/tag/v8.0.0

.github/workflows/nightly.yml
jokasimr: Fixed!

YooSunYoung (approved):

Issue #70, opened 2026-02-10
NeXus workflow: Why the name `FrameMonitor`?


nvaytet:
Stupid question: where did the name "frame monitor" come from?
In the nexus types, we have `FrameMonitor0-3`, but a more common name (at least when talking to scientists) seems to be "beam monitor".

Should we rename `FrameMonitor` to `BeamMonitor`?

jl-wynen:
It comes from BIFROST. I think we can remove it here since we are now able to define custom monitors in the subpackages.

nvaytet:
Does this mean we should also remove things like `ScatteringRunType` (which doesn't seem to be used in any downstream packages: https://github.com/search?q=org%3Ascipp%20ScatteringRunType&type=code).

And maybe also:
- `EmptyBeamRun` (from sans)
- `VanadiumRun` (from powder)
- `TransmissionRun` (from sans)
- `CaveMonitor` (from dream)
- `IncidentMonitor` (from sans)
- `TransmissionMonitor` (from sans)
?

jl-wynen:
This needs to be a stepwise procedure:
1. Define these types in the technique packages and release.
2. Then remove the types from ESSreduce.

PR #607, opened 2026-05-28
Move types to subpackages

jl-wynen:
First step of #70

### DIFF
--- a/packages/essreduce/src/ess/reduce/nexus/types.py
+++ b/packages/essreduce/src/ess/reduce/nexus/types.py
@@ -103,20 +103,11 @@ ESSreduce provides the following but custom types can be used:
 - :class:`VanadiumRun`
 """
 
-
 # 1.2  Monitor types
 IncidentMonitor = NewType('IncidentMonitor', int)
 """Incident monitor"""
 TransmissionMonitor = NewType('TransmissionMonitor', int)
 """Transmission monitor"""
-FrameMonitor0 = NewType('FrameMonitor0', int)
-"""Frame monitor number 0"""
-FrameMonitor1 = NewType('FrameMonitor1', int)
-"""Frame monitor number 1"""
-FrameMonitor2 = NewType('FrameMonitor2', int)
-"""Frame monitor number 2"""
-FrameMonitor3 = NewType('FrameMonitor3', int)
-"""Frame monitor number 3"""
 CaveMonitor = NewType('CaveMonitor', int)
 """A monitor located in the instrument cave"""
 MonitorType = TypeVar('MonitorType')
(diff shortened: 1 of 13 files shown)


nvaytet (commented):
It looks good but I am worried the changes are going to clash a lot with #603 😬 
Which one should we merge first?

jl-wynen:
Depends on how close that PR is to merging? Resolving conflicts should be simple

nvaytet:
My guess is that it will still take a while to figure everything out correctly in that PR.
So that answers my question...

packages/essreduce/src/ess/reduce/nexus/types.py
```diff
@@ -103,40 +106,24 @@ class TransmissionRun(Generic[ScatteringRunType]):
 - :class:`VanadiumRun`
 """
 
-
 # 1.2  Monitor types
-IncidentMonitor = NewType('IncidentMonitor', int)
-"""Incident monitor"""
-TransmissionMonitor = NewType('TransmissionMonitor', int)
-"""Transmission monitor"""
-FrameMonitor0 = NewType('FrameMonitor0', int)
```
nvaytet, line 112: Would there be any use in providing a default set of monitors? (`Monitor1-4`?)

packages/essreduce/src/ess/reduce/nexus/types.py
```diff
@@ -58,6 +58,11 @@ class ProductionInfo(snx.NXsource):
 like ``SampleRun`` or ``BackgroundRun``.
 """
 
+SampleRun = NewType('SampleRun', int)
```
nvaytet, line 61: Why is the `SampleRun` special?
Shouldn't we remove all in the future, and sub-packages can decide if they want to name it `SampleRun` or something else?

Is there any code internally to `essreduce` that makes use of `SampleRun` or is it only in the tests/docs?

packages/essreduce/src/ess/reduce/nexus/types.py
jl-wynen: I don't think so. We shouldn't use them in workflows because they carry no meaning. So I think they would only be good for testing, and then we can easily define them in the tests.

packages/essreduce/src/ess/reduce/nexus/types.py
jl-wynen: It is special in how I and maybe we think about workflows. But that is not a good argument on its own.

There are some uses in `live` and `polarization`. Removing the ones in `live` might be tricky. But maybe that code should be in `esslivedata` anyway? And for polarization, there is this: #615

nvaytet: I would have thought that everything in `essreduce` could just be `RunType`, unless I am forgetting something?

> But maybe that code should be in esslivedata anyway?

Yes, I would vote to move it there.

jl-wynen: @SimonHeybrock ?

SimonHeybrock: I think every instrument does sample-runs, so having a unified name makes sense (or else every instrument will have their own naming convention)?

In ESSlivedata I actually just used this as a default in some cases. Defining `AnyRun` would be more thruthful. But there are probably also uses via technique packages.

nvaytet: > or else every instrument will have their own naming convention

They seem to already do that with `BackgroundRun`, `OpenBeamRun`, `EmptyRun`, `DirectRun`... so maybe leaving the freedom to define a sample run is not so different?

nvaytet: So was the conclusion here to keep `SampleRun` and move forwards with the changes as they are?

jl-wynen: I think so.

nvaytet (approved):

PR #614, opened 2026-05-28
[CI] ci: generate badges from nightly job matrix

jokasimr:
In our old multi-repo setup each badge on the ess-project page was associated with a single workflow run.

In the current setup that is not the case, for two reasons:

1. All three nightly tests are run in the same workflow (this could be changed).
2. Each package is tested in the same workflow (this could probably not be changed).

This means that if we want "per package" badges in the status page, we cannot use the `/badge.svg` url associated with the workflow, like we did before.

This PR implements a solution for this problem that lets us have a single nightly workflow for all nightly types and all packages, but it requires us to generate the badge svgs ourselves based on results retrieved from the github api in a separate step in the workflow.

It does sound a bit icky, but I think having per-project badges on the status page might be worth it.
What do you think? Is there a better solution?

Alternatives:

1. Split the nightly workflows per nightly-type (main, latest, lowest) and only display one badge per nightly-type, not one per package.
2. Create one separate workflow for each nightly-type / package combination.

### DIFF
--- a/.github/scripts/generate-nightly-badges.py
+++ b/.github/scripts/generate-nightly-badges.py
@@ -1,0 +1,213 @@
+#!/usr/bin/env python3
+"""Generate package-level status badges for the nightly workflow."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import urllib.error
+import urllib.request
+from pathlib import Path
+from typing import Any
+from xml.sax.saxutils import escape
+
+PACKAGES = (
+    "essreduce",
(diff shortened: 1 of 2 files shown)


YooSunYoung (approved):
I looks reasonable.
We've been collecting dev-tools in the `tools` folder.
Should we keep the scripts in the same place maybe?

jokasimr:
I think for now we'll just have a single badge for all test runs. We can always add this later if we feel it would be useful.

PR #617, opened 2026-06-01
[ESSDIFFRACTION] Implement more PDF functions

### LINKED ISSUE #468: [Requirement] PDF : support other expression of output functions
### Executive summary

Provide choice of other output functions from PDF data reduction

### Context and background knowledge

It is quite common to plot different quantities at the end of PDF data reduction.

- Pair correlation function $G(r) = \frac{2}{\pi}\int_0^{Q_{\rm max}} Q \; [S(Q)-1] \; \sin(Qr) \; {\rm d}Q$
- Pair distribution function $g(r) = 1+ \frac{G(r)}{4\pi \rho r },$ where $\rho$ is the atomic density
- Radial distribution function ${\rm RDF}(r) = 4\pi \rho r^2 g(r),$ where $\rho$ is the atomic density
- Linearised radial distribution function $T(r) = {\rm RDF}(r)/r$
- Running coordination number $ \int_0^r {\rm RDF}(r') \; {\rm d}r'$

The first one is already implemented in essdiffraction: `ess.powder.transform.compute_pdf_from_structure_factor`

The follow-up work is to add this final step to the PDF data reduction workflow in essdiffraction (not done yet)

### Inputs

Input: S(Q)

### Methodology

See above for the expressions of the formula to implement

### Outputs

Output: different 2-column files where the first one is r and the other is one of the following quantities: G(r), g(r), T(r), RDF(r) or the running coordination number.

### Which interfaces are required?

Python module / function

### Test cases

See attached zipped folder (jupyter notebook and test input and output datafiles

[PDF_Scipp_requirement_files_for_tests.zip](https://github.com/user-attachments/files/25002666/PDF_Scipp_requirement_files_for_tests.zip)

### Existing implementations

_No response_

### Comments

_No response_

jl-wynen:
Fixes #468

Draft because error propagation is missing. But can you take a look anyway because I am unsure about the approach of requiring the user to call a chain of these functions instead of, e.g., always computing the final result from S(r). 

I moved the existing implementation to `ess.diffraction` because it is not directly related to powder.

Does this need more tests? The functions are quite simple and tricky to test.


jokasimr (commented):
>  I am unsure about the approach of requiring the user to call a chain of these functions instead of, e.g., always computing the final result from S(r).

I'm not sure either. But I think we can keep it like this for now, because using these functions it's very easy to add functions that lets the user compute any of the quantities directly from `S(r)`. 

Most users would even be able to do that themselves.

packages/essdiffraction/src/ess/diffraction/pdf.py
```diff
@@ -0,0 +1,276 @@
...
+        cov_g = _covariance_of_matrix_vector_product(c * v, ioq)
+        cov_g = sc.DataArray(
+            cov_g, coords={d: r.rename_dims({r.dim: d}) for d in cov_g.dims}
+        )
+        return g, cov_g
+    return g
+
+
+def pair_distribution_function(
+    g: sc.DataArray, *, atomic_density: sc.Variable
```
jokasimr: Should the argument name be capital `G` instead of `g`?

And similar in the return value description of `pair_correlation_function`?

In the docs it looks like `g` is the input and `g` is the output.

jokasimr:
> Does this need more tests? The functions are quite simple and tricky to test.

I don't think they need more tests. Like you said the functions are quite simple, and they already have some tests.

If we really want to lock down the behavior we could have a test case where we pass a constant `G(r)` to the other functions and check what their outputs are what we would expect in that case.

packages/essdiffraction/src/ess/diffraction/pdf.py
jl-wynen: I renamed it

jl-wynen:
Any more comments?

jokasimr (approved):

PR #621, opened 2026-06-03
[ESSSPECTROSCOPY] Default BIFROST detector names

jl-wynen:
There is no need to load them from the file any more because they no longer contain an index and so are now fixed. Before, the detector names could change when components were added or removed from the files.

### DIFF
--- a/packages/essspectroscopy/src/ess/bifrost/detector.py
+++ b/packages/essspectroscopy/src/ess/bifrost/detector.py
@@ -282,7 +282,7 @@ def merge_triplets(
     return concatenated
 
 
-def detector_names() -> list[str]:
+def default_detector_names() -> list[str]:
     """Return the list of detector names in a BIFROST NeXus file.
 
     These names are based on the names in the NeXus schema.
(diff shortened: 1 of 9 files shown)


packages/essspectroscopy/src/ess/bifrost/detector.py
```diff
@@ -281,4 +282,17 @@ def merge_triplets(
     return concatenated
 
 
+def detector_names() -> list[str]:
```
nvaytet: I forget now what we do for the monitor names? Where do we define them?

nvaytet: Can we call this function `default_detector_names` (a bit like `default_parameters()`)?
I got confused when reading the changes as I thought you were passing this function to the `BifrostWorkflow` but then seemingly never calling it anywhere (a bit like a pytest fixture).

Until I finally found that it is imported as
```Py
from .detector import detector_names as default_detector_names
```

So the argument in `BifrostWorkflow` is just a list that could be passed by the user to customize the names.

jl-wynen: They are needed for mapping the workflow over the banks.

nvaytet (approved):

PR #623, opened 2026-06-04
[CI] Do not wait for Tests to build Docs in CI

nvaytet:
Resoning:

Failing early is useful, but these days the unit tests seem to be almost as slow as building the docs.

It is sometimes useful to see that both tests and docs builds are broken, so we can fix everything at once.

Also, if the author of the PR has done things properly and tested both tests and docs locally, it shouldn't really fail.

It does seem a bit strange to have to wait for the long unit tests to all complete while we could spend some of that waiting time starting to build docs, since we have quite a few runners for the organisation? (this is especially true when pushing changes to `essreduce`).

Total runtime for running everything decreased from 22m to ~14m~ :slightly_smiling_face: 

Edit: got it down to 10m30s by not executing the notebooks during doctest and linkcheck

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -92,7 +92,7 @@ essreduce = { path = "packages/essreduce", editable = true, extras = ["test", "d
 args = [
     { "arg" = "builder", "default" = "html" },
 ]
-cmd = "python -m sphinx -v -b {{ builder }} -d packages/essreduce/.docs_doctrees packages/essreduce/docs packages/essreduce/html"
+cmd = "python -m sphinx -v -b {{ builder }}{% if builder in ['linkcheck', 'doctest'] %} -D nbsphinx_execute=never{% endif %} -d packages/essreduce/.docs_doctrees packages/essreduce/docs packages/essreduce/html"
 default-environment = "docs-essreduce"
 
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -92,7 +92,7 @@ essreduce = { path = "packages/essreduce", editable = true, extras = ["test", "d
 args = [
     { "arg" = "builder", "default" = "html" },
 ]
-cmd = "python -m sphinx -v -b {{ builder }} -d packages/essreduce/.docs_doctrees packages/essreduce/docs packages/essreduce/html"
+cmd = "python -m sphinx -v -b {{ builder }}{% if builder in ['linkcheck', 'doctest'] %} -D nbsphinx_execute=never{% endif %} -d packages/essreduce/.docs_doctrees packages/essreduce/docs packages/essreduce/html"
```
jl-wynen, line 95: Didn't know this supports these advanced template features. :+1:

jl-wynen (approved):

pixi.toml
nvaytet: I didn't know either, but apparently copilot does!

PR #630, opened 2026-06-09
[ESSREDUCE] Project chopper axle onto incident beam for flight distance

SimonHeybrock:
## Problem

The chopper-cascade wavelength LUT derives each chopper's flight distance from `norm(axle_position - source_position)`. A disk chopper's axle sits above, below, or to the side of the beam by design, so this straight-line distance overestimates the actual beam-path length.

Usually the error is negligible, but for closely-spaced choppers a large transverse axle offset can reorder the cascade. Concrete case on DREAM: the pulse-shaping choppers PSC1 (axle 0.7 m off-beam, 6.145 m downstream) and PSC2 (on-beam, 6.155 m). Under the norm their distances become 6.185 m and 6.155 m, so PSC1 is sorted *after* PSC2 and applied last — or, when the LUT is evaluated over a narrow distance range around the choppers, never applied at all. The result is that PSC1's phase/delay has no effect on the table.

## Fix

Compute the distance as the projection of the axle position onto the incident-beam direction. The neutron crosses the disk where the incident beam intersects it, so this is the correct flight distance. Assuming the incident beam runs along the z axis, the projection is simply the z-component of the offset from the source. This is consistent with the straight-line `Ltotal` used for detectors and monitors, which lie on the beam where the projection reduces to the norm, so on-axis choppers are unaffected and existing results are unchanged.

Taking the z-component (rather than a source-to-sample dot product) avoids needing a sample position as input, keeping the workflow API unchanged.

## Tests

- Unit test that an off-beam axle projects onto the beam.
- Regression test that two closely-spaced choppers with one offset axle keep the correct cascade order (the DREAM PSC1/PSC2 situation).

Existing analytical and simulation LUT tests pass unchanged.

### DIFF
--- a/packages/essreduce/src/ess/reduce/unwrap/lut.py
+++ b/packages/essreduce/src/ess/reduce/unwrap/lut.py
@@ -429,9 +429,44 @@ def _to_component_reading(component) -> BeamlineComponentReading:
     )
 
 
+def chopper_distance_along_beam(
+    axle_position: sc.Variable,
+    source_position: sc.Variable,
+    sample_position: sc.Variable,
+) -> sc.Variable:
+    """Flight distance from the source to a chopper along the incident beam.
+
+    A disk chopper's axle is offset from the beam (the disk sits above or below
+    it), so the straight-line distance ``norm(axle_position - source_position)``
+    overestimates the flight distance, and for closely-spaced choppers can even
+    place them in the wrong order along the cascade. The neutron crosses the disk
+    where the incident beam (source to sample) intersects it, so the flight
+    distance is the projection of the axle position onto the incident-beam
+    direction. This matches the straight-line ``Ltotal`` used for detectors and
+    monitors, where the component lies on the beam and the projection reduces to
+    the norm.
+
+    Parameters
+    ----------
+    axle_position:
+        Position of the chopper's rotation axle.
+    source_position:
+        Position of the neutron source.
+    sample_position:
+        Position of the sample, defining the incident-beam direction together
+        with the source.
+    """
+    source_position = source_position.to(unit=axle_position.unit)
+    incident_beam = sample_position.to(unit=axle_position.unit) - source_position
+    return sc.dot(
+        axle_position - source_position, incident_beam / sc.norm(incident_beam)
+    )
+
+
 def simulate_chopper_cascade_using_tof(
     choppers: DiskChoppers[RunType],
     source_position: Position[snx.NXsource, RunType],
+    sample_position: Position[snx.NXsample, RunType],
     neutrons: NumberOfSimulatedNeutrons,
     pulse_stride: PulseStride[RunType],
     seed: SimulationSeed,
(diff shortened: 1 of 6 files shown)


packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -429,9 +429,44 @@ def _to_component_reading(component) -> BeamlineComponentReading:
...
+    incident_beam = sample_position.to(unit=axle_position.unit) - source_position
+    return sc.dot(
+        axle_position - source_position, incident_beam / sc.norm(incident_beam)
+    )
+
+
 def simulate_chopper_cascade_using_tof(
     choppers: DiskChoppers[RunType],
     source_position: Position[snx.NXsource, RunType],
+    sample_position: Position[snx.NXsample, RunType],
```
SimonHeybrock: @nvaytet Not sure about this, as it would require always having a sample-position. Alternative would be to simply default to beam=Z.

packages/essreduce/src/ess/reduce/unwrap/lut.py
```diff
@@ -429,9 +429,44 @@ def _to_component_reading(component) -> BeamlineComponentReading:
...
+        Position of the chopper's rotation axle.
+    source_position:
+        Position of the neutron source.
+    sample_position:
+        Position of the sample, defining the incident-beam direction together
+        with the source.
+    """
+    source_position = source_position.to(unit=axle_position.unit)
+    incident_beam = sample_position.to(unit=axle_position.unit) - source_position
+    return sc.dot(
```
nvaytet: I had assumed that for balancing purposes, choppers are always installed vertical?

In the case of a tilted guide, I think the dot product is not quite right, because the chopper disk goes vertically from its axle center towards the beam, instead of perpendicular to the beam?
Instead we may need to somehow use the gravity vector?

Image

We also need to make it work for the case where the chopper is neither above nor below the beam, but on the side (as is the case for one of the Dream choppers).

SimonHeybrock: This comment seems to go beyond the immediate fix -- the old solution didn't handle curved guides either, did it? Or do you think it is worse than before?

nvaytet: > the old solution didn't handle curved guides either, did it? Or do you think it is worse than before?

No and no. It's just a case I thought about when reading your use of the dot product.
I think we should either fix this additional case, or at least write a comment in the code and/or open an issue?

nvaytet: Conclusion from in-person discussion: use the z-component, which also removes the need for having a sample position.

SimonHeybrock: @nvaytet Done, ready for review

nvaytet (approved):

PR #631, opened 2026-06-09
[essimaging] Use on-the-fly wavelength lookup table for Odin

nvaytet:
Move the Odin Bragg edge workflow to build the wavelength lookup table on-the-fly from chopper parameters in the file.

For now, we patch the chopper settings on the workflow, as the info is incomplete in the template coda file that was used to create the test files.

Needs https://github.com/scipp/scippneutron/pull/707


packages/essreduce/src/ess/reduce/nexus/workflow.py
```diff
@@ -570,13 +575,24 @@ def to_disk_choppers(choppers: RawChoppers[RunType]) -> DiskChoppers[RunType]:
     choppers:
         Raw choppers loaded from the NeXus file, as a DataGroup.
     """
-    disk_choppers = {
+    disk_choppers = {}
+    for key, ch in choppers.items():
+        if (
```
nvaytet, line 580: If a chopper has some empty logs, it is dropped as it is considered parked/not in use.

packages/essdiffraction/pyproject.toml
```diff
@@ -29,16 +29,10 @@ requires-python = ">=3.11"
 # Run 'tox -e deps' after making changes here. This will update requirement files.
 # Make sure to list one dependency per line.
 dependencies = [
-  "dask>=2022.1.0",
```
nvaytet, line 32: All these dependencies are brought in by `essreduce`.
Removing them here avoids the issue of always keeping all the lower bounds up to date (right now, most of the lower bounds we had were lying anyway).

jl-wynen: I would still list all direct dependencies here. Or at least direct 3rd party dependencies (dask, graphviz) to make sure we don't rely on essreduce's setup. We don't need to list lower bounds here as long as they are not tighter than essreduce's bounds.

packages/essreduce/src/ess/reduce/nexus/workflow.py
jl-wynen: Do you know how this will be handled in production? Will nexus files omit parked choppers or will they always be there? We shouldn't emit a warning if length-0 logs are a normal thing.

nvaytet: My assumption is that the choppers will always be there, because they are in the template files used by ecdc.
But I asked now in the ecdc channel.

> We shouldn't emit a warning if length-0 logs are a normal thing.

True. Let's see what ecdc say and then we can probably remove the warning.

nvaytet: No reply from ECDC so far.
I removed the warning for now, as I am quite sure parked choppers will be in the files.

jl-wynen (approved):

PR #633, opened 2026-06-10
[ESSREDUCE] Generalize cylinder mantle projection to axis-aligned cylinders

SimonHeybrock:
## Motivation

scipp/esslivedata#970 needs to add projections for MAGIC, which has detectors like DREAM mantle but along a vertical cylinder.

## Overview

Generalizes the cylinder-mantle projection in `live.raw` from a hard-coded z-axis cylinder to any axis-aligned cylinder, motivated by the need for a cylinder-Y projection (vertically-aligned mantle detectors). Only axis-aligned variants are supported, which is sufficient for current detectors.

`project_onto_cylinder_z` becomes `project_onto_cylinder` with an `axis` parameter ('x', 'y', or 'z'), using a right-handed cyclic relabeling of the in-plane axes so the math is shared. `from_nexus` accepts the new `'cylinder_mantle_y'` projection alongside the existing `'cylinder_mantle_z'`.

To avoid forcing a synchronized release with downstream (esslivedata calls `make_cylinder_mantle_coords` directly), that function keeps its name and signature, gaining only an `axis` keyword defaulting to `'z'`. Existing direct calls are unchanged; cylinder-Y is opt-in via `axis='y'`. The `from_nexus` projection strings are likewise additive.

The pixel-noise path is untouched: it derives from physical pixel-tube geometry, independent of the projection axis.

Note: for `'cylinder_mantle_y'` the `resolution` dict must key the axial dimension `'y'` (rather than `'z'`), since the output coordinate name follows the cylinder axis.

## Test plan

- New unit test for the cylinder-Y projection, mirroring the existing z test under axis relabeling.
- Existing z test updated to the new signature.
- Verified esslivedata's `projectors_test.py` still passes against this branch (direct `make_cylinder_mantle_coords(coords)` call preserved).


packages/essreduce/tests/live/raw_test.py
```diff
@@ -274,6 +267,29 @@ def test_project_onto_cylinder_z() -> None:
     )
 
 
+def test_project_onto_cylinder_y_is_z_result_under_axis_relabeling() -> None:
```
nvaytet: Also add a test for `x` case, as it is an option above?

SimonHeybrock: Good call -- added in 2a68cca0a, mirroring the y/z tests under (y, z) in-plane relabeling.

packages/essreduce/tests/live/raw_test.py
```diff
@@ -290,6 +290,29 @@ def test_project_onto_cylinder_y_is_z_result_under_axis_relabeling() -> None:
     )
 
 
+def test_project_onto_cylinder_x_is_z_result_under_axis_relabeling() -> None:
```
nvaytet: How about parametrizing the test?

SimonHeybrock: Done!

nvaytet (approved):

PR #635, opened 2026-06-10
[CI] Add release note generating step in the release action

YooSunYoung:
Release note editting steps are added in the `release.yml`.

It's split into two jobs:
- download all the relevant commits/pr information using github API (bash script)
- generate the release note from the downloaded information (python script)

I tested in a fork https://github.com/YooSunYoung/ess/releases/tag/essnmx%2F26.6.2
but not sure if it would work as expected in the action here as well...

# Manual Test
Here is the example how I ran the scripts using `essnmx/26.6.0` and `essnmx/26.4.0` as an example.

## Prerequisites
```
git fetch --tags  # Or only relevant tags...
```
## Download the commit information
The script only runs with `bash` since it has some syntax that `sh` doesn't have...
```
bash tools/download-release-diff-info.sh essnmx/26.6.0 essnmx/26.4.0
```
## Generate release note.
```
pixi run --frozen python tools/collect-release-notes.py --cur-tag essnmx/26.6.0 --compare-tag essnmx/26.4.0
```

By default, the generated note will be stored under `.ess_release_cache/`, <br>
so in this case, `.ess_release_cache/essnmx-26.6.0-essnmx-26.4.0-releasenote.md`.

Here is the example output from this test:

### ESSnmx

* Bump essreduce to 26.6.0 by @jl-wynen in https://api.github.com/repos/scipp/ess/pulls/624
* [ESSNMX] Save TOF 1d histogram by @YooSunYoung in https://api.github.com/repos/scipp/ess/pulls/600
* ci: add lower bound and latest released nightlies by @jokasimr in https://api.github.com/repos/scipp/ess/pulls/606
* [NMX] Reference data comparison test. by @YooSunYoung in https://api.github.com/repos/scipp/ess/pulls/546
* [NMX] Allow setting output time bin unit. by @YooSunYoung in https://api.github.com/repos/scipp/ess/pulls/545
* [Essnmx] Use time bin width instead of number of bin edges. by @YooSunYoung in https://api.github.com/repos/scipp/ess/pulls/264

**Full Changelog**: https://github.com/scipp/ess/compare/essnmx/26.4.0...essnmx/26.6.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,16 @@ on:
   push:
     tags:
       - "*/[0-9]*"
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Release tag - hardcoded.'
+        required: true
+        type: string
+      dry-run:
+        description: 'Dry run the release action.'
+        default: false
+        type: boolean
 
 env:
   PIXI_FROZEN: true
@@ -19,7 +29,11 @@ jobs:
     steps:
       - id: parse
         run: |
-          TAG="${GITHUB_REF#refs/tags/}"
+          if [ ${{ inputs.dry-run }} = true ]; then
+            TAG="${GITHUB_REF#refs/tags/}"
+          else
+            TAG=${{ inputs.tag }}
+          fi
           PACKAGE="${TAG%%/*}"
           VERSION="${TAG#*/}"
           echo "package=${PACKAGE}" >> "$GITHUB_OUTPUT"
@@ -50,6 +64,7 @@ jobs:
     needs: [ determine-package, build ]
     runs-on: ubuntu-24.04
     environment: release
+    if: ${{ !inputs.dry-run }}
     permissions:
       id-token: write
     steps:
@@ -69,7 +84,7 @@ jobs:
     uses: ./.github/workflows/docs.yml
     with:
       package: ${{ needs.determine-package.outputs.package }}
-      publish: true
+      publish: ${{ !inputs.dry-run }}
       linkcheck: false
 
   assets:
@@ -91,6 +106,7 @@ jobs:
           zip -r documentation-${PACKAGE}-${VERSION}.zip documentation-${PACKAGE}-${VERSION}
       - name: Upload release assets
         uses: svenstaro/upload-release-action@v2
+        if: ${{ !inputs.dry-run }}
         with:
           file: ./documentation-${{ needs.determine-package.outputs.package }}-${{ needs.determine-package.outputs.version }}.zip
           overwrite: false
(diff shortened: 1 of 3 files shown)


jokasimr:
Interesting! Will this workflow overwrite whatever was in the release notes originally? Or just append to it?

YooSunYoung:
> Interesting! Will this workflow overwrite whatever was in the release notes originally? Or just append to it?

I just used `edit`, which will simply overwrite whatever there was if any.
I assumed that there was nothing yet when this action is triggered.

jokasimr:
If possible I think it would be better to append to the existing release notes. Then the person making the release can just leave them empty and they'll be populated automatically, or they can add their own release notes and the generated notes will be appended below.

.github/workflows/release.yml
```diff
@@ -94,3 +94,75 @@ jobs:
...
+        run: ls .ess_release_cache
+      - name: Generate release note
+        run: python tools/collect-release-notes.py --cur-tag ${PACKAGE}/${VERSION} --compare-tag ${PACKAGE}/${LAST_VERSION}
+      - name: Merge existing release note with generated note
+        run: |
+          gh release view ${PACKAGE}/${VERSION} -R scipp/ess --json body -q .body >> existing-note.md
+          if [ -s existing-note.md ]; then
+            cat existing-note.md >> new-release-note.md
+          fi
+          cat .ess_release_cache/${PACKAGE}-${VERSION}-${PACKAGE}-${LAST_VERSION}-releasenote.md >> new-release-note.md
```
YooSunYoung: @jokasimr I added this step to handle the existing release note.
Is that what you meant in the comment...?

i.e. in case someone wrote something in advance in the release note.

jokasimr: Yes that's what I meant. 

Is there a reason why the code appends to `existing-note.md`, then checks if it's empty, then writes it to `new-release-note.md` and then appends the generated note to `new-release-note.md`?

Can we not write the existing release note to `new-release-note.md` directly and then append the generated note to `new-release-note.md`?

jokasimr (commented):
Looks good to me! It would be very useful to have this.

How do we test this?
Would it help to add a "dry-run" option that runs without actually replacing the description of the release?
Or are there better options?

.github/workflows/release.yml
YooSunYoung: > Can we not write the existing release note to new-release-note.md directly and then append the generated note to new-release-note.md?

I don't know how to avoid appending a new line if the body is empty in a clean way.

YooSunYoung:
> Looks good to me! It would be very useful to have this.
> 
> How do we test this? Would it help to add a "dry-run" option that runs without actually replacing the description of the release? Or are there better options?

~Ah maybe we should use the publish flag. I'll update it...~
No... that was `docs` workflow... 
But we could have a dry-run flag with some input arguments to the action.
I think it'll make release workflow debugging much easier.

YooSunYoung:
I implemented the dry-run mode with hard-coded tag input argument.
It has to be merged to main first, in order to use the manual workflow_dispatch in the github action page.

So... should we merge it first and then test it with the github workflow_dispatch event trigger on dry-run mode...?

.github/workflows/release.yml
jokasimr: Aha I see, that makes sense :+1:

jokasimr:
> So... should we merge it first and then test it with the github workflow_dispatch event trigger on dry-run mode...?

Sounds reasonable to me!

jokasimr (approved):

PR #639, opened 2026-06-17
Bump plopp version to avoid deprecation warning at the lowest bound.

YooSunYoung:
There was a deprecation warning in the nightly lower bound tests: https://github.com/scipp/ess/actions/runs/27670291862/job/81832902677

I could fix it by bumping up plopp version.

essnmx doesn't have instrument view helper yet but set it anyway...


.github/workflows/nightly.yml
YooSunYoung, line 1: It looked like the setup-uv was using the same hash for both lower bound and latest so it was warning about the cache conflicts.

packages/essreflectometry/pyproject.toml
```diff
@@ -45,9 +45,10 @@ dynamic = ["version"]
 test = [
     "pytest>=7.0",
     "pooch>=1.5",
+	"ipywidgets>=8.1.7",
 ]
 gui = [
-    "ipywidgets>=8.0.0",
+	"ipywidgets>=8.1.7",
```
YooSunYoung: It wasn't in the test environment so it raised in my workstation but it didn't complain in the nightly... I didn't look into more detailed but it looks like it pulls some extra stuff??? 

I added the ipywidgets with the same version as essdiffraction.

YooSunYoung:
I tested the fixed nightly here: https://github.com/scipp/ess/actions/runs/27686771078/job/81889224537

Ignore the essnmx latest nightly test since it'll be fixed when the essreduce is released...

pixi.lock
YooSunYoung, line 1: I had to update the pixi lock file as well.

packages/essreflectometry/pyproject.toml
nvaytet: Which test was failing on your workstation? I'm curious as to why we need ipywidgets for the tests...

YooSunYoung: `tests/reflectometry/figures_test.py` had failing tests where it fails to build figures, using helper functions such as `wavelength_z_figure`.

nvaytet: I don't see anything in the source or the tests that would require either widgets or interactive figures (unless somehow you have interactive figures enabled by default?)

YooSunYoung: Oh maybe... yeah it didn't fail in the nightly so maybe we can just remove this dependency... I'll remove it...

packages/essreflectometry/pyproject.toml
```diff
@@ -45,9 +45,10 @@ dynamic = ["version"]
 test = [
     "pytest>=7.0",
     "pooch>=1.5",
+	"ipywidgets>=8.1.7",
```
YooSunYoung:
```suggestion
```

packages/essreflectometry/pyproject.toml
YooSunYoung: @nvaytet I removed it as it seems like it's my workstation environment problem...

I skipped the CI cause it shouldn't affect any of the regular tests.
Can you bypass the rule and send it to the merge queue unless there's anything else to fix?

nvaytet (approved):

PR #640, opened 2026-06-17
[CI] Add dependency bumping tool.

YooSunYoung:
While fixing the lower-bound environment, it was useful to have a script that updates dependencies of multiple packages at once....

I used it for #639 

The intended way of using is like this:
```bash
pixi run --frozen python tools/bump-dependency-version.py \
  --package ess* \
  --dependency-name plopp \
  --version 26.5.0
```

Then it'll update all packages' `plopp` minimum version to 26.5.0.
It ignores if `plopp` is not there dependency.

or combined with running tests...
```
pixi run --frozen python tools/bump-dependency-version.py \
  --package essdiffraction \
  --dependency-name plopp \
  --version 26.4.2

cd packages/essdiffraction && uv run --extra=test --resolution=lowest-direct pytest && cd -
```

Then I could easily try different version of dependency to find the proper lowest-direct version of the dependency.
I wrote the script to allow patterns to update multiple packages' dependencies

### DIFF
--- a/tools/bump-dependency-version.py
+++ b/tools/bump-dependency-version.py
@@ -6,6 +6,40 @@ import re
 
 import tomllib
 
+"""Minimum dependency updating script.
+
+This script updates minimum dependency to the given version
+if any matching package depends on it(dependency-name).
+
+Example 1 - update minimum version of plopp to 26.5.0 for all packages:
+```
+pixi run --frozen python tools/bump-dependency-version.py \
+  --package ess* \
+  --dependency-name plopp \
+  --version 26.5.0
+```
+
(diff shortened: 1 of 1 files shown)


tools/bump-dependency-version.py
```diff
@@ -0,0 +1,141 @@
+import argparse
```
MridulS, line 1: Can we add some in line docs to this file? Like how to run
```
pixi run --frozen python tools/bump-dependency-version.py \
  --package ess* \
  --dependency-name plopp \
  --version 26.5.0
```

MridulS (approved):

PR #641, opened 2026-06-17
[essimaging] Publish reduced scitiff from Odin bragg edge workflow

nvaytet:
We add a unit test that publishes a normalized image as a scitiff file.
This is to be picked up by analysis team for further tests.

The would like to have `tof` as the 3rd dimension, so we add a provider to compute it.


nvaytet:
Needs https://github.com/scipp/scitiff/issues/56

packages/essimaging/tests/odin/data_reduction_test.py
```diff
@@ -64,3 +70,46 @@ def test_can_compute_wavelength(run_type, wavelength_mode):
...
+    res_openbeam = tools.resample(openbeam, sizes=new_sizes)
+    den = res_openbeam.hist(tof=tbins)
+
+    normed = num / den
+
+    to_scitiff = (
+        normed.assign_coords(
+            x=normed.coords['x_pixel_offset'],
+            y=normed.coords['y_pixel_offset'],
+            t=sc.midpoints(normed.coords['tof']),
```
YooSunYoung: I think we should keep the coordinate names as they are, no...?

YooSunYoung (approved):
Looks good to me except for the minor comment about the coordinate names in the output file.

packages/essimaging/tests/odin/data_reduction_test.py
nvaytet: I just assumed that people reading the images expect to have x/y as coordinates instead of our strange 'pixel_offset' coords?

nvaytet: Also, we rename the dims below, so it sort of makes sense that the coords have the same name as the dims? But maybe I misunderstood the Scitiff data structure?

YooSunYoung: The dimension names are limited so that we make sure of the order of those dimensions.
It's because, imaging processing SW such as ImageJ often assumes certain order of dimensions.

Coordinate names should stay informative so that people know what they actually mean on the other hand.
However, it doesn't matter too much since they will have units.... I guess...?

Like, `t` dimension can have either `tof` or `wavelength` but unit will tell you which is which.
It'll be more problematic if they have same units like `tof` and `event_time_offset`, but this is less likely to happen...?

About the `x/y_pixel_offset`. I think it is important to keep them as they are. It'll be less confusing even.
For example, x_pixel_offset doesn't represent its actual x-position of pixel in a certain coordinate.
x_pixel_offset is clear that it's only referring to the detector plane.

PR #644, opened 2026-06-18
[CI] Add draft release step to GitHub Actions workflow

YooSunYoung:
The release note upload job failed because the release was not drafted/published yet.
It was depending on the assets uploading job to create the release if not present.

It used default values for creating the release if there is None for the given tag.
Then the title was set to a weird commit message instead of the tag name (it depends on how many commits belong to the PR apparently?).

So I added an explicit release draft step with the specific title(tag name).
In this way we can use the runners a bit more efficiently... (release note generating and docs build can take a while.)

Also I removed the upload file action and used `gh` cli directly instead.

The release note uploading job is responsible for publishing the release,
i.e. the release is set to latest/published from draft when there is a release note.

Assets uploading (built docs) won't be necessarily done by the time the release is published.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,6 +25,7 @@ jobs:
     outputs:
       package: ${{ steps.parse.outputs.package }}
       version: ${{ steps.parse.outputs.version }}
+      tag: ${{ steps.parse.outputs.tag }}
       lastversion: ${{ steps.last-version-parse.outputs.lastversion }}
     env:
       DRYRUN: ${{ github.event_name == 'workflow_dispatch' }}
@@ -39,6 +40,7 @@ jobs:
           fi
           PACKAGE="${TAG%%/*}"
           VERSION="${TAG#*/}"
+          echo "tag=${TAG}" >> "$GITHUB_OUTPUT"
           echo "package=${PACKAGE}" >> "$GITHUB_OUTPUT"
           echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
       - uses: actions/checkout@v6
@@ -99,6 +101,19 @@ jobs:
         with:
           packages-dir: dist/
 
+  draft-release:
+    name: Publish release
+    needs: [ publish, determine-package ]
+    runs-on: ubuntu-slim
+    permissions:
+      contents: write
+    steps:
+      - name: Draft a release.
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ needs.determine-package.outputs.tag }}
+        run: gh release create ${TAG} --draft --title ${TAG}  # Default title is not always the tag name.
+
   docs:
     name: Build and publish docs
     needs: [ determine-package, publish ]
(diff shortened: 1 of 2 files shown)


YooSunYoung:
Sorry for pushing things continuously... it's now really ready...!

.github/workflows/release.yml
```diff
@@ -99,6 +99,19 @@ jobs:
         with:
           packages-dir: dist/
 
+  draft-release:
```
nvaytet, line 102: This step should be green even if creating the release fails.
For example, if release was created via the web interface.

.github/workflows/release.yml
```diff
@@ -128,14 +141,15 @@ jobs:
...
       - name: Upload release assets
-        uses: svenstaro/upload-release-action@v2
-        with:
-          file: ./documentation-${{ needs.determine-package.outputs.package }}-${{ needs.determine-package.outputs.version }}.zip
-          overwrite: false
+        env:
+          GH_TOKEN: ${{ github.token }}
+          ASSET-FILE-PATH: ./documentation-${{ needs.determine-package.outputs.package }}-${{ needs.determine-package.outputs.version }}.zip
+          TAG: ${{ needs.determine-package.outputs.package }}/${{ needs.determine-package.outputs.version }}
+        run: gh release upload ${TAG} ${ASSET-FILE-PATH} --clobber  # Overwrite existing assets of the same name.
```
nvaytet, line 152: Unrelated to these changes: can we use the same (get rid of the `svenstaro/upload-release-action`) for uploading nightly reduced files here: https://github.com/scipp/ess/blob/main/.github/workflows/nightly.yml#L90

.github/workflows/release.yml
YooSunYoung: I made the step success optional and then added a checking step

.github/workflows/release.yml
YooSunYoung: I think so ...?

YooSunYoung:
I manually updated the release note btw: https://github.com/scipp/ess/releases/tag/essreduce%2F26.6.2

.github/workflows/release.yml
nvaytet:
```suggestion
    needs: build
```

I don't think we need the `determine-package` here as we don't use any of its outputs?

tools/collect-release-notes.py
```diff
@@ -107,12 +107,7 @@ def has_label(label: str, pr: PRDescription) -> bool:
         else:
             maybe_relevant_logs.append(merge_log)
 
-    release_note = [
-        "## What's Changed",
-        '',
-        f"### {package_name.replace('ess', 'ESS')}",
```
nvaytet, line 113: Can you explain why this is now gone?

YooSunYoung: It's already included in the release title so it was unnecessarily verbose.

Or we can integrate that into the other section title..? Like:

## What's Changed in ESSreduce/26.6.2

nvaytet (approved):

PR #646, opened 2026-06-22
[ESSNMX] Update essnmx reduced file for regression test

YooSunYoung:
I had to update the frozen reduced file due to changes in the `tof`.
I added change log of these files in the developers guide too.

This update should fix the failing nightly(latest) test.

### DIFF
--- a/packages/essnmx/src/ess/nmx/_nxlauetof_io.py
+++ b/packages/essnmx/src/ess/nmx/_nxlauetof_io.py
@@ -188,7 +188,8 @@ def load_essnmx_nxlauetof(file: str | FilePath | NeXusFile) -> sc.DataGroup:
             dg = f[()]
 
         # Drop auxiliary output from the loaded result.
-        del dg['entry']['aux']
+        if 'aux' in dg['entry']:
+            del dg['entry']['aux']
         _validate_entry(entry := f['entry'])
         _handle_sample(dg['entry']['sample'], entry['sample'])
         _handle_monitor(dg['entry']['control'], entry['control'])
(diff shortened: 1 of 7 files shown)


packages/essnmx/src/ess/nmx/_nxlauetof_io.py
```diff
@@ -188,7 +188,8 @@ def load_essnmx_nxlauetof(file: str | FilePath | NeXusFile) -> sc.DataGroup:
             dg = f[()]
 
         # Drop auxiliary output from the loaded result.
-        del dg['entry']['aux']
+        if 'aux' in dg['entry']:
+            del dg['entry']['aux']
```
YooSunYoung, line 192: It was a bug that needed to be fixed for the documentation.

packages/essnmx/tools/freeze-reduced.py
```diff
@@ -0,0 +1,33 @@
...
+)
+from ess.nmx.executables import reduction
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--freeze-version",
+        help="Future version of essnmx that can reproduce the frozen file.",
+        required=True,
+    )
```
SimonHeybrock, line 46: The description (e.g., inline help) for this script is a bit thin, can it be extended? Unclear when it should be used, and how it related to the `test-dataset.ipynb`.

SimonHeybrock (commented):
Just a question about docstrings/help.

SimonHeybrock (approved):

PR #647, opened 2026-06-22
[ESSDIFFRACTION] feat: add workflow using lookup tables for pulse shaping modes

### LINKED ISSUE #628: [essdiffraction, BEER] Lookup table support for pulse shaping modes
BEER will support 2 categories of chopper configurations: modulation and pulse shaping.

Lookuptables could be used for the pulse shaping modes.

In McStas and for Diffraction, these modes are modes 4, 5, 6.

Refer to the commissioning notebook stored in [Code Shelf](https://git.esss.dk/dram/code-shelf/code-shelf/-/tree/main/src/beer/notebooks/commissioning?ref_type=heads) to get the corresponding chopper settings.

There are a [few NeXus files using Si powder as sample](https://github.com/scipp/ess/blob/main/packages/essdiffraction/src/ess/beer/data.py). Note that these files might correspond to an old model of the detectors.

jokasimr:
Fixes #628 

Chopper information taken from https://git.esss.dk/dram/code-shelf/code-shelf/-/blob/main/src/beer/notebooks/commissioning/beer_choppers_tof.py?ref_type=heads


packages/essdiffraction/src/ess/beer/beamline.py
```diff
@@ -0,0 +1,198 @@
...
+    "FC2A": {
+        "frequency": -14.0,
+        "phase": -133.67285314925246,
+        "distance": 79.975,
+        "open": [0.0],
+        "close": [175.0],
+    },
+}
+
+parameters: dict[PulseShapingMode, dict[str, _ChopperParameters]] = {
```
jl-wynen:
```suggestion
_parameters: dict[PulseShapingMode, dict[str, _ChopperParameters]] = {
```
Since they are defined in terms of a protected name (`_ChopperParameters`) and don't have units.

packages/essdiffraction/src/ess/beer/beamline.py
```diff
@@ -0,0 +1,198 @@
...
+        },
+    },
+}
+
+
+def choppers(
+    mode: PulseShapingMode, source_position: sc.Variable
+) -> MappingProxyType[str, DiskChopper]:
+    """
+    List of ESS BEER choppers for the selected pulse-shaping mode.
```
jl-wynen:
```suggestion
    Dict of ESS BEER choppers for the selected pulse-shaping mode.
```

packages/essdiffraction/src/ess/beer/mcstas/beamline.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def choppers(
+    mode: PulseShapingMode, source_position: sc.Variable
+) -> MappingProxyType[str, DiskChopper]:
+    """
+    List of ESS BEER choppers for the selected pulse-shaping mode.
+
+    We make the chopper information available in this way as loading it directly from
+    the NeXus files is currently not available for these simulated BEER data.
```
jl-wynen, line 178: Should the function have a name that clarifies this to avoid people accidentally using it over chopper info from files? Something like `simulation_choppers` or `default_choppers`?

packages/essdiffraction/src/ess/beer/conversions.py
```diff
@@ -207,7 +207,7 @@ def t0_estimate(
         * source_to_wavelength_definition_chopper_distance.to(
             unit=wavelength_estimate.unit
         )
-    ).to(unit='s')
+    ).to(unit='s') + sc.scalar(2.86, unit='ms').to(unit='s') / 2
```
jl-wynen: What is this number? And why did you add it now?

packages/essdiffraction/src/ess/beer/io.py
```diff
@@ -1,504 +1,13 @@
...
-
-
-mcstas_providers = (
-    load_beer_mcstas_provider,
-    load_beer_mcstas_monitor_provider,
-    mcstas_chopper_delay_from_mode,
-)
+def __getattr__(name: str):
+    """Return attributes from the replacement :mod:`ess.beer.mcstas` module."""
+    return getattr(_mcstas, name)
```
jl-wynen, line 21: Can you raise a deprecation warning?

packages/essdiffraction/src/ess/beer/mcstas/load.py
jl-wynen, line 1: Did anything change in this file? The diff isn't usable.

packages/essdiffraction/src/ess/beer/conversions.py
jokasimr: Because I noticed that in the beer McStas model the `t` event data value (time of detection) is modified by a offset value that is half a ess pulse width. This makes the source pulse centered around `t=0` in the McStas model. To convert that `t` value to `event_time_offset` with the convention that we expect (`event_time_offset=0` corresponds to the start of a pulse) I now add the same offset to `event_time_offset` when loading McStas data. And to keep the results unchanged we have to shift the time here by the same offset.

I think this makes more sense than the old version. The `t0_estimate` is supposed to be the time a "nominal" wavelength neutron reaches the wavelength defining chopper. But that time is not defined unless we also specify a "nominal" time the neutron was created at the source. Before this change that time was `0`, now it is `source_pulse_width / 2`, this is more in line with the conventions I'm used to.

packages/essdiffraction/src/ess/beer/mcstas/load.py
jokasimr: Two things changed:

1. In the `load_beer_mcstas` function the definition of `event_time_offset` changed. Now a small offset of `pulse_width/2` is added, so that we have the convention that the source pulse starts at `event_time_offset=0`. The `frame_cutoff_time` was also shifted correspondingly.
2. New providers needed for the lookup table frame unwrapping workflows are added after line 530 in the file.

packages/essdiffraction/src/ess/beer/conversions.py
jl-wynen: Ok. Can you add this explanation to the code? It's hard to find in a GH comment.

packages/essdiffraction/src/ess/beer/mcstas.py
```diff
@@ -346,6 +346,12 @@ def _load_beer_mcstas(f, *, north_or_south=None, number=None, detector_sizes):
     )
 
     pulse_width = sc.scalar(2.86, unit='ms').to(unit='s')
+    # The `t` value in the Beer model McStas file is offset by
+    # half a pulse width (so that the pulse is centered around t=0).
+    # But that is different from the convention that we have for
+    # `event_time_offset`. We expect `event_time_offset=0` to correspond
+    # to the start of a pulse.
+    # To fix this, half a pulse width is added to `t`.
```
jl-wynen: Please also add a short comment (maybe with a link to this one) in the other place where you add the offset.

jokasimr: Nah I don't think that is necessary because the logic there is completely natural when we work with the standard convention that the start of the pulse is `event_time_offset=0`.

jl-wynen: But it uses a 'magic' number. I'd like to at least have an explanation  what that number represents.

jl-wynen (approved):

PR #648, opened 2026-06-23
[ESSIMAGING] Upgrade scitiff for essimaging reduced file.

YooSunYoung:
Unblocks #641


pixi.toml
nvaytet: This is the `conda` section.
Either remove this section or add `scitiff` there also?

nvaytet (approved):

PR #650, opened 2026-06-25
[ESSSANS] [ESSDIFFRACTION] deps: use plopp[all]

jokasimr:
Fixes https://github.com/scipp/ess/actions/runs/28149207702/job/83362844781
https://github.com/scipp/ess/actions/runs/28149207702/job/83362844775

We are using some functionality in esssans and essdiffraction that requires the extended functionality of plopp. 
Instead of surgically adding all required dependencies here, I think it makes sense to just depend on `plopp[all]`.


packages/essdiffraction/pyproject.toml
```diff
@@ -33,7 +33,7 @@ dependencies = [
   "graphviz>=0.20",
   "essreduce>=26.6.0",
   "numpy>=2",
-  "plopp>=26.5.0",
+  "plopp[all]>=26.5.0",
```
nvaytet: I guess you probably need to update the `pixi.lock` file as well?

SimonHeybrock:
This means more dependencies for `esslivedata` (or any auto/batch reduction service you implement), which is not using plotting at all. Maybe not a true issue, but not a nice approach either?

jokasimr:
> This means more dependencies for esslivedata

Do you prefer adding only the necessary packages (no `all`). Or do you prefer adding them only to the test and docs extras?

SimonHeybrock:
> > This means more dependencies for esslivedata
> 
> Do you prefer adding only the necessary packages (no `all`). Or do you prefer adding them only to the test and docs extras?

Conclusion from chat: We have said in the past that instrument-packages might be considered "apps", i.e., get all dependencies a user might need. However, that may have forgotten other package use (livedata,auto-reduction,batch-reduction). If we add other GUI-related dependencies on top of plotting (such as widget libraries), we suddenly pull in a lot more than needed. -> Will continue discussion in person.

jl-wynen:
Someone just mentioned in person to put plopp into the test dependencies only. I think that batch usages will (almost) always want plotting capabilities to make diagnostic plots alongside the reduced data. So I think we should keep at least static plotting dependencies in the defaults.

jokasimr:
> Someone just mentioned in person to put plopp into the test dependencies only. I think that batch usages will (almost) always want plotting capabilities to make diagnostic plots alongside the reduced data. So I think we should keep at least static plotting dependencies in the defaults.

Yeah I also think that making the dependencies excessively "lean" to avoid running into hypothetical dependency problems/issues in the future is not the right way to go.

Excluding some heavy dependencies and hiding them behind extras is fine, but overall I think we avoid more issues if we try to make our packages "batteries included" even if that means pulling in some dependencies that are not used most of the time.

packages/esssans/pyproject.toml
nvaytet: I see esssans has pythreejs as a hard dep.
Should we move it to test deps?

jokasimr: Maybe, but unless that can be answered directly by someone I think that is a separate discussion that should not block this fix for the nightlies?

nvaytet: I don't think any user has noticed or even cares? For now, Oliver probably has a large env with everything installed or just uses VISA. Same for Judith.

So this is more for us. I think we just haven't noticed before since packages were in separate repos. Now that we have everything in one place, we notice differences better. I would vote to move it to test, to make is consistent with the other packages.

`pythreejs` is a dep for the same reason as the other packages: the instrument view.

jokasimr: I'll move it to test.

nvaytet (approved):

PR #654, opened 2026-06-26
refactor: use parameter models from esslivedata

### LINKED ISSUE #653: Consolidate workflow specifications, parameter models, and metadata mechanisms
## Motivation

Avoid duplication, unify descriptions, etc. for UI generation and user info.

Relevant bits to be consolidated

- https://github.com/scipp/ess/blob/main/packages/essnmx/src/ess/nmx/configurations.py
- https://github.com/scipp/esslivedata/blob/main/src/ess/livedata/parameter_models.py
- Trimmed-down version of https://github.com/scipp/esslivedata/blob/aaf21dcf92a4df4e2b8b1e7278da96e02cd647b0/src/ess/livedata/config/workflow_spec.py#L246-L252
Probably not for v1:
- Instrument-specific metadata (https://github.com/scipp/esslivedata/blob/aaf21dcf92a4df4e2b8b1e7278da96e02cd647b0/src/ess/livedata/config/instruments/loki/specs.py#L182-L227 )
- Models for inputs and outputs (example: https://github.com/scipp/esslivedata/blob/aaf21dcf92a4df4e2b8b1e7278da96e02cd647b0/src/ess/livedata/config/instruments/loki/specs.py#L120-L170)
Maybe start by writing an ADR about a minimal solution so we can have a consensus about scope and shape?

## Mandatory reading

I have previously done most of this analysis in scipp/esslivedata#889. Make sure to read it and in particular https://github.com/scipp/esslivedata/issues/889#issuecomment-4311180193 -- the attempt to consolidate everything ended up getting out of hand and blocked the entire work.

jokasimr:
Fixes #653 

This is a codex rewrite of the parameter / workflow widget system using the models from esslivedata.


packages/essdiffraction/src/ess/beer/workflow.py
```diff
@@ -12,41 +12,67 @@
...
+        | (two_theta > sc.scalar(float('-inf'), unit=two_theta.unit))
+    ),
+    TofMask: None,
+    WavelengthMask: None,
+    MaskedDetectorIDs: MaskedDetectorIDs({}),
+    NeXusDetectorName: "detector",
+    DetectorBank: DetectorBank.south,
+    Filename[SampleRun]: str(mcstas_powder_silicon_in_vanadium_can()),
+    Filename[VanadiumRun]: str(mcstas_powder_vanadium()),
+    Filename[EmptyCanRun]: str(mcstas_powder_empty_can()),
```
jokasimr, line 65: These changes will be revisited or removed.

They were added mainly for debugging purposes.

packages/esssans/src/ess/sans/parameters.py
```diff
@@ -51,71 +50,207 @@
...
+    stop: float = Field(default=0.5, description="Stop of the Qy edges.")
+
+    def get_edges(self) -> sc.Variable:
+        return sc.linspace(
+            dim='Qy',
+            start=self.start,
+            stop=self.stop,
+            num=self.num_bins + 1,
+            unit=self.unit.value,
+        )
```
jokasimr, line 83: Why does this have to be re-implemented here?
Looks like only the `dim` and descriptions should need to be specified.

packages/esssans/src/ess/sans/workflow.py
```diff
@@ -141,13 +158,15 @@ def with_background_runs(
     return _set_runs(workflow, runs, BackgroundRun, 'background_run')
 
 
-parameter_mappers[PixelMaskFilename] = with_pixel_mask_filenames
-# TODO: for now, we leave the mapping over detector banks out, because we do not have a
-# method to merge the I(Q) of different banks, and we thus cannot compute a single
-# result from the workflow. So only a single detector bank can be processed at a time.
-# parameter_mappers[NeXusDetectorName] = with_banks
```
jokasimr, line 148: Comment should not be removed.

packages/essreduce/src/ess/reduce/workflow.py
```diff
@@ -3,45 +3,160 @@
...
+
+
+@dataclass(frozen=True)
+class WorkflowSpec:
+    """Metadata needed to generate interfaces for a workflow."""
+
+    name: str
+    title: str | None = None
+    description: str | None = None
+    version: str | None = None
```
SimonHeybrock, line 35: Why does the `WorkflowSpec` not hold models for workflow inputs and outputs, or the param-specs you defined?

packages/essreduce/src/ess/reduce/parameter.py
```diff
@@ -21,158 +20,53 @@ class KeepDefaultType:
...
+    model: Any
+    category: str
+    title: str | None = None
+    description: str | None = None
+    default: T | KeepDefaultType = keep_default
+    transform: Callable[[T], Any] | None = None
+    apply: Callable[[Pipeline, T], Pipeline] | None = None
+    filter_keys: tuple[Key, ...] = ()
+    use_workflow_default: bool = True
+    key: Key | None = None
```
SimonHeybrock, line 35: This seems like a central part we should talk about. It seems like the idea is that a set of `ParameterSpec` would/could take the role of the input-params-model (and output-params-model)? One concern I have is that this ties the spec to the implementation (`transforms,` `key`, ...), which would force full dependencies in UI packages that do not run code. `title`, `description`, and `default` we'd get for free as field attributes if we used a model holding all params. On the other hand, storing them in this param spec seems like it has the advantage in composability, as we can provide multiple workflows with different subsets of param specs (exposed inputs).

packages/essdiffraction/src/ess/dream/parameters.py
```diff
@@ -1,73 +1,124 @@
...
+parameters[UncertaintyBroadcastMode] = ParameterSpec(
+    model=UncertaintyBroadcastMode,
+    category='Reduction',
+    title='Uncertainty Broadcast',
+    default=UncertaintyBroadcastMode.upper_bound,
 )
 
-# TODO: the mask parameters (TofMask, TwoThetaMask, WavelengthMask) need a new widget
-# that allows to define a python function in the notebook and pass it to the workflow.
-# We defer this to later; the masks are set to None by default in the workflow for now.
```
jokasimr, line 70: Comment should be preserved.

packages/essreduce/src/ess/reduce/parameter.py
jokasimr: I think there are some suspect fields that were added by codex for quite hacky reasons, for example `filter_keys`. I'll try to iterate a bit on this to make it leaner. 

But I don't really understand what you mean, how could the input params model not hold a reference to the sciline `Key` that it intends to represent? Or do you mean that the input parameters is another separate layer of input parameters that are then mapped to sciline keys in a separate stage? Or did you mean something different?

It's probably best to talk in chat. Do you have time at 13?

SimonHeybrock: In ESSlivedata the workflow factory is responsible for this. Only it know the implementation and how to map model fields to workflow keys (there may even be non 1:1 mappings), the model does not.

packages/essreduce/src/ess/reduce/workflow.py
jokasimr: Updated it so that now it does. I agree that makes more sense.

jokasimr:
Replaced by https://github.com/scipp/ess/pull/690

PR #663, opened 2026-07-01
[essimaging] Orca normalization by proton charge per frame

### LINKED ISSUE #636: [essimaging] Orca image normalization was minsunderstood
After discussion with the Odin instrument scientist, some things were clarified about normalizing Orca images.

### Exposure time

We probably do not need to know the camera exposure time after all. The assumption is that for every run, the dark frame, open beam and sample will all be recorded with the same FPS on the camera. So the exposure time factor should cancel out everywhere.

We could still read the exposure time as a safety check, to make sure we are not combining data that were recorded using different FPS.

### Normalizing by proton charge

Currently, the implementation of the proton charge normalization is wrong. We need to normalize each recorded frame by the corresponding proton charge.

For every single frame (before we do any mean or sum of the dark/open frames), we need to find the pulse that corresponds to the `time` trigger coordinate in the orca 2d array, and use that to normalize the frame. Only once this is done can we average dark and open beam frames together.

### Recording modes

There seems to be two main recording modes for the camera:

- Start/stop: most probably synchronized with a pulse trigger; rotate sample > take a frame > rotate sample again > take new frame
- Continuous: goes out of sync with the pulse at either 30, 60 or 100 fps?; rotate continuously and record at high frame rate

What do we do for frames that end up between pulses? It is currently not clear what the data will even look like in the nexus files, as currently the time axis is controlled by the pulses. How will it look when wer record continuously at 100 fps?

To get some more information on this, the Odin team will carry out some tests with the orca camera in the coming weeks.

nvaytet:
We change the way we use the proton charge to normalize the Orca data.

- We normalize each frame in the Dark, OpenBeam, and Sample runs by the corresponding proton charge.
- After that we take the mean of the Dark along the `time` dimension.
- We then remove the MeanDark from the OpenBeam and Sample data.
- Finally, we divide the `(Sample - MeanDark) / (OpenBeam - MeanDark).mean('time')`

Fixes #636


packages/essimaging/src/ess/tbl/data.py
```diff
@@ -8,16 +8,16 @@
...
     files={
         "tbl_sample_data_2025-03.hdf": "md5:12db6bc06721278b3abe47992eac3e77",
         "TBL-wavelength-lookup-table-no-choppers-5m-35m.h5": "md5:e28793b7e1c12986ee63a1df68723268",  # noqa: E501
         'tbl-orca-focussing.hdf.zip': Entry(
             alg='md5', chk='f365acd9ea45dd205c0b9398d163cfa4', unzip=True
         ),
-        "ymir_lego_dark_run.hdf": "md5:c0ed089dd7663986042e29fb47514130",
-        "ymir_lego_openbeam_run.hdf": "md5:00375becd54d2ed3be096413dc30883c",
-        "ymir_lego_sample_run.hdf": "md5:ae56a335cf3d4e87ef090ec4e51da69c",
+        "ymir_lego_dark_run.hdf": "md5:2d9ae7d6f1d0502c17b0030bcb3dad1d",
```
nvaytet, line 18: I had to update the data because the proton charge in the old files started exactly at the start of the time axis for the camera data. However, when doing `sc.lookup` in 'previous' mode, the first data point was not found (probably because of numerical precision errors) and a NaN was returned.

So the first data slice in the dark lookup was NaN. We then took the mean which gave NaN, and then all data was NaN.
I now updated the proton charge logs to start a bit before the data starts.

packages/essimaging/src/ess/tbl/orca.py
```diff
@@ -28,117 +22,45 @@
...
-        data.sum('time') / charge_per_frame.sum('time')
-    )
-
+    # A note on timings:
+    # We want to sum the proton charge inside each time bin (defined by the duration of
+    # each frame). However, the time dimension of the data recorded at the detector is
+    # not the same time as the proton charge (which is when the protons hit the
+    # target). We need to shift the proton charge time to account for the time it takes
+    # for neutrons to travel from the target to the detector. Does this mean we cannot
+    # do the normalization without computing time of flight?
```
jl-wynen, line 48: Yes, I think that is what this implies. Or is ODIN short enough that we can just use the 'time' coord from the nexus data?

What is the 'time' coord in `data` here? Is that the time at the source?

packages/essimaging/src/ess/imaging/normalization.py
```diff
@@ -42,7 +57,8 @@ def subtract_background_sample(
 
 def subtract_background_openbeam(
     data: FluxNormalizedDetector[OpenBeamRun],
-    background: FluxNormalizedDetector[DarkBackgroundRun],
+    background: MeanDarkFrame,
```
nvaytet, line 61: Note: Should we subtract the mean dark from each OB frame or should we subtract it from the mean OB? Are the resulting variances the same?

packages/essimaging/src/ess/imaging/normalization.py
```diff
@@ -10,16 +10,31 @@
...
+    MeanDarkFrame,
     NormalizedImage,
     OpenBeamRun,
     SampleRun,
     UncertaintyBroadcastMode,
 )
 
 
+def average_dark_frames(
+    dark_frames: FluxNormalizedDetector[DarkBackgroundRun],
```
nvaytet: Should we actually normalize the dark measurement by proton charge? If it never sees the neutrons, we could be introducing fluctuations that aren't actually there?

nvaytet: If there is background from e.g. a neighouring instrument bleeding in through the cave walls, then fluctuations could be related to the proton charge after all?

-> assume to begin with that we don't need to normalize by proton charge but keep the option open to add it back in later.

packages/essimaging/src/ess/imaging/normalization.py
jl-wynen: It's also a question about the scales of float values. Are there significant truncation errors from subtracting the (large?) mean dark from each individual OB frame?

packages/essimaging/src/ess/imaging/normalization.py
jl-wynen: Sounds to me like you don't want to normalise by proton charge but by measurement duration. You need something that corresponds to an event rate, not a count. The proton charge norm does this but it depends on the source. So if you would measure a dark frame while the source is off, it wouldn't count under proton charge norm. But that makes no sense.

packages/essimaging/src/ess/imaging/normalization.py
nvaytet: I think we can safely assume that counts in the dark frame will always be lower than in open beam and sample. Open beam will probably have the highest counts.

packages/essimaging/src/ess/imaging/normalization.py
nvaytet: For the measurement duration, the conclusion was that all 3 measurements will be recorded with the same frame rate (fps). So every frame has the exact same duration. That duration cancels out in the normalization so we actually don't need the exposure time of each frame.

packages/essimaging/src/ess/imaging/normalization.py
jl-wynen: But is `sum(open_beam) >> sum(dark_frame)`? In that case, you *might* get better accuracy by subtracting frame by frame.

packages/essimaging/src/ess/imaging/normalization.py
jl-wynen: For my understanding: Does this mean that every frame is normalised by the duration such that `sum(dark_frame)` does not depend on the number of dark frames recorded?

packages/essimaging/src/ess/imaging/normalization.py
nvaytet: I also think we expect the dark frame to be more stable than the open beam. The dark is the inherent noise of the chip which is always there. The open beam can fluctuate due to external factors. So I would vote to subtract the dark from each OB frame and then take the mean OB after that.

packages/essimaging/src/ess/imaging/normalization.py
nvaytet: We don't take the sum of dark frames, we take the mean. The same for the open beam.
I think as long as every frame has been normalized by the corresponding proton charge, and everything is in units of `counts/C` (or something like that), then this should work fine?

jl-wynen: Sounds ok. But I don't know the underlying equation you're solving here.

nvaytet: So if we decide not to normalize the dark frames by the proton charge, we have a units issue.
The sample frame is in counts/C while the dark frame is just counts.

I think this implies that we need to subtract the dark frame first (as it should never depend on proton charge) and then normalize the dark-subtracted sample and OB runs by proton charge?

jl-wynen: Good point

jl-wynen:
The latest changes look good. But what about https://github.com/scipp/ess/pull/663#discussion_r3517660261 ?

packages/essimaging/src/ess/tbl/orca.py
nvaytet: The comment here was already present in the code (in the `_compute_proton_charge_per_exposure` function). This is not new, and I still have not figured out how to resolve this.
Fixing it should not be part of this PR.

nvaytet:
> The latest changes look good. But what about [#663 (comment)](https://github.com/scipp/ess/pull/663#discussion_r3517660261) ?

See reply

jl-wynen (approved):

PR #664, opened 2026-07-02
[essimaging] Odin image normalization wf

nvaytet:
We add the analog to the TBL Orca image normalization for Odin.

Odin will not have 3 separate files for dark, open beam and sample runs, all will be recorded in one file.
The data is split according to `ImageKey` (as was for Ymir).

We are able to re-use the orca workflow from TBL (that was moved into the common `imaging` folder) and insert providers that spit the data. We use a new RunType `AllRuns` to load a single file first, and then use a trick to re-use the `FluxNormalizedDetector` to transform `AllRuns` into `SampleRun, `OpenBeamRun`, and `DarkBackgroundRun`.

The visualization graph looks a little weird but the workflow works fine:
Screenshot_20260702_221825

Note that this is a little different from the Ymir implementation.

### Edit:

I had to change the workflow graph because now we subtract the Dark background **before** we normalize by proton charge. So instead we use image key at the `CorrectedDetector` step, and we use the full `ProtonCharge[AllRuns]` log when normalizing by proton charge. It doesn't really hurt to search inside a slightly longer log, and this was simpler than also extracting the correct part of the log for each file (as we would have to make sure we get all the relevant data, also in the future taking into account travel time down the beam path).

The graph now looks like
Screenshot_20260709_155946


YooSunYoung (approved):
LGTM...!

We probably have to support merging multiple files too.
e.g. Users Acquired SampleRun, OpenBeam over night and then later decided to collect more OpenBeam in another file.
But I don't have a good idea now...

YooSunYoung (approved):
It looks good...!

PR #666, opened 2026-07-03
[CI] Remove draft release CI step

nvaytet:
The draft release step when making a release goes wrong when we make a release via the web UI.

A draft release with the same name using the same tag is created alongside the release already created by the web UI.
Screenshot_20260703_132021

Then, the release notes of the non-draft release get updated.
The uploading of the docs as assets fails.
The draft release never gets promoted to non-draft, because the original release gets picked up by the `gh` command.

For now, we remove this step, meaning that releases should only be created via the web interface (simply pushing a tag from the CLI will no longer work).
We should revisit once everyone is back from vacation.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -99,24 +99,6 @@ jobs:
         with:
           packages-dir: dist/
 
-  draft-release:
-    name: Draft a release or/and check the release
-    needs: [ determine-package ]
-    runs-on: ubuntu-slim
-    if: ${{ startsWith(github.ref, 'refs/tags/') && github.event_name != 'workflow_dispatch' }}
-    permissions:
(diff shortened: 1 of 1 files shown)


jl-wynen (approved):

YooSunYoung:
The `gh release create` changed its behaviour ... it used to fail when there is a release with the same tag already but now it goes through. We could probably fix it by adding an if condition in the draft release creating step but it's also fine we just stick to the `gh` client and web UI...??

One downside is just we'll have to type more things when we want to run batch releases.
With the `tag` way, we just need to push multiple tags but with this change we'll have to use `gh release`... which is not too bad....??

But I think the docs build assets upload failed for different reason. I'll try fixing it...

PR #671, opened 2026-07-23
[ESSSPECTROSCOPY] Update BIFROST NeXus loader for angles

jl-wynen:
The integration tests are failing because of this. It looks like the CODA files encode the angles we need.

I am unhappy with relying on group names this much but I don't see a better alternative given that there are no unique NeXus classes to look for.


jl-wynen:
Should I add test data for this? It would be a fairly large file with bogus data. So it wouldn't be useful beyond testing this loader.

packages/essspectroscopy/src/ess/bifrost/io/mcstas.py
```diff
@@ -7,15 +7,46 @@
...
+
+def _load_experiment_parameter(
+    file_spec: NeXusFileSpec[RunType], param_name: str
+) -> sc.DataArray:
+    with open_component_group(
+        NeXusLocationSpec(filename=file_spec.value),
+        nx_class=snx.NXparameters,
+        parent_class=snx.NXentry,
+    ) as group:
+        return group[param_name][()]['value']
```
SimonHeybrock, line 39: So McStas stores this differently than PROD?

packages/essspectroscopy/src/ess/bifrost/io/nexus.py
```diff
@@ -25,33 +25,57 @@
...
+
+        if transform.transformation_type != 'rotation':
+            raise ValueError(
+                "Expected the instrument angle at detector_tank_angle to be a rotation,"
+                f" got '{transform.transformation_type}' instead.'"
+            )
+        return transform.value  # type: ignore[return-value]
+
+
+# See https://github.com/scipp/essreduce/issues/98
```
SimonHeybrock: Should this link to an `scipp/ess` issue instead?

packages/essspectroscopy/src/ess/bifrost/io/nexus.py
```diff
@@ -25,33 +25,57 @@
...
+        nx_class=snx.NXinstrument,
         parent_class=snx.NXentry,
-    ) as group:
-        return group[param_name][()]['value']
+    ) as instrument:
+        log: snx.Group = instrument[path]  # type: ignore[assignment]
+        if isinstance(log['value'], snx.Group):
+            # The NXlog is nested in this group, e.g., the group can be an NXpositioner
+            # with children 'value', 'target_value', etc.
+            transform: snx.nxtransformations.Transform = log['value'][()]  # type: ignore[assignment]
```
SimonHeybrock, line 62: Still trying to understand whether this kind of nesting is a bug/bad and should be reported. Have had issues with this elsewhere. But no action here, as I presume it is necessary to make it work now.

packages/essspectroscopy/src/ess/bifrost/io/mcstas.py
jl-wynen: The McStas files that Greg made have an `NXparameters` group which doesn't exist in prod because it holds simulation params. The handling of nested value groups in the prod loader can load those data from mcstas files. Unfortunately, the sample stack does not exist in mcstas files. But the instrument angle can be loaded. So I could remove its special handler for mcstas. Should I?

SimonHeybrock: Maybe?

packages/essspectroscopy/src/ess/bifrost/io/nexus.py
jl-wynen: Yes, it wouldn't work otherwise.

SimonHeybrock (approved):

PR #673, opened 2026-07-23
[ESSdiffraction] Save XYE files in tests and apply masks

### LINKED ISSUE #618: [essdiffraction]: Publish reduced files in xye format
Currently we are only publishing CIF files.
XYE files are more standard; these files could then be used to test other software (GSAS, etc) in the integration tests.

We can just use the same data.

jl-wynen:
Fixes #618 and #672

The XYE files do not contain any metadata (see https://github.com/scipp/scippneutron/issues/706) but they should be good enough to run some integration tests.

### DIFF
--- a/packages/essdiffraction/tests/dream/geant4_reduction_test.py
+++ b/packages/essdiffraction/tests/dream/geant4_reduction_test.py
@@ -10,6 +10,7 @@ import sciline
 import scipp as sc
 import scipp.testing
 from ess import dream, powder
+from ess.dream.io import save_xye
 from ess.dream.workflows import (
     DreamGeant4MonitorHistogramWorkflow,
     DreamGeant4MonitorIntegratedWorkflow,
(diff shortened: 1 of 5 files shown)


packages/essdiffraction/tests/dream/geant4_reduction_test.py
```diff
@@ -302,6 +305,8 @@ def test_pipeline_save_data_to_disk(output_folder: Path):
     """
     result.save(output_folder / "dream_reduced.cif")
 
+    wf.bind_and_call(save_xye)
```
nvaytet, line 308: Is the reason why we don't use `bind_and_call` for CIF files because we want to add the `result.comment`?

At first glance it seems strange that we use one method for CIF file and another for XYE file.

nvaytet (approved):

packages/essdiffraction/tests/dream/geant4_reduction_test.py
jl-wynen: We could handle the comment differently. And having a free-form comment for XYE or other files could be useful, too. The difference comes mainly from there being this `CIF` object that you build in memory. So the workflow has something to return. But we don't have that for other formats. (We do for SQW but that is way more complicated.) 
We can talk about a good, common interface if you like.

PR #674, opened 2026-07-24
[ESSDIFFRACTION] Fix multiplication with mask

jl-wynen:
I messed this up in #673

- ~Zero out masked elements, not unmasked ones~ keep unmasked elements
- ~Apply mask to variances~
- ~use `sc.where` instead of multiplication because the latter does not work when the data is Inf or NaN.~
- Add test to check written data

**Update:** analysis team suggested that masked values should simply be removed from the file written to disk, instead of replacing with zeros or NaNs.
This has the advantage of circumventing the issue with what to do with the variances.

### DIFF
--- a/packages/essdiffraction/src/ess/dream/io/_common.py
+++ b/packages/essdiffraction/src/ess/dream/io/_common.py
@@ -14,13 +14,8 @@ def prepare_reduced_data(da: sc.DataArray) -> sc.DataArray:
     hist.coords[hist.dim] = sc.midpoints(hist.coords[hist.dim])
 
     if (mask := irreducible_mask(hist.masks, hist.dim)) is not None:
-        # No file format we use here supports masks, so the next
-        # best thing is to zero out masked data:
-        if hist.variances is not None:
-            replacement = sc.scalar(0.0, variance=0.0, unit=hist.unit, dtype=hist.dtype)
-        else:
-            replacement = sc.scalar(0.0, unit=hist.unit, dtype=hist.dtype)
-        hist.data = sc.where(mask, replacement, hist.data)
+        # No file format we use here supports masks, so we remove masked data points
+        hist = hist[~mask]
         hist.masks.clear()
 
     return hist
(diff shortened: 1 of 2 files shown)


packages/essdiffraction/src/ess/dream/io/_common.py
```diff
@@ -13,11 +13,14 @@ def prepare_reduced_data(da: sc.DataArray) -> sc.DataArray:
...
     hist.coords[hist.dim] = sc.midpoints(hist.coords[hist.dim])
 
-    if hist.masks:
+    if (mask := irreducible_mask(hist.masks, hist.dim)) is not None:
         # No file format we use here supports masks, so the next
         # best thing is to zero out masked data:
-        hist.data = hist.data.copy()
-        hist.values *= irreducible_mask(hist.masks, hist.dim).values
+        if hist.variances is not None:
+            replacement = sc.scalar(0.0, variance=0.0, unit=hist.unit, dtype=hist.dtype)
```
nvaytet: I guess this is the usual question: should the variance be 0, and thus being a very tight constraint on fits?

nvaytet: I think I would nan the masked values instead of 0. Makes them easier to distinguish from real zeros

jl-wynen: Good point. I tried that and easydiffraction doesn't like it. It raises
```
TypeError: Value mismatch for <reduced_tof.data.1.intensity_meas>. Provided nan outside [0, inf]. Keeping current 0.0.
```
Judging by that error message, `inf` would be valid.

@AndrewSazonov How do we best encode masked data? In this case, it is masked because the vanadium measurement has a 0 in a bin.

nvaytet: Inf feels like it would just be a workaround/hack. Would make more sense if NaN was accepted downstream.

nvaytet (approved):

PR #684, opened 2026-07-29
[ESSNMX] Fix weekly test for NMX

YooSunYoung:
It was just because of the bitshuffle.h5 ...

I tested manually https://github.com/scipp/ess/actions/runs/30463403880

The other issue was that the plotting was done with interactive backend in the windows.
I just set the default backend to be `Agg` for NMX executable since it only dumps PNG files.
It's already a known issue: https://github.com/scipp/plopp/issues/516

### DIFF
--- a/packages/essnmx/tests/executable_output_test.py
+++ b/packages/essnmx/tests/executable_output_test.py
@@ -31,7 +31,10 @@ try:
     # so that h5py can use the plugin.
     import bitshuffle.h5  # noqa: F401
 except ImportError:
-    ...
+    pytest.skip(
+        allow_module_level=True,
+        reason="File output regression test cannot be done without `bitshuffle.h5`.",
+    )
 
 
 def assert_h5_attrs_equal(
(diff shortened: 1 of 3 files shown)


packages/essnmx/tests/executable_output_test.py
```diff
@@ -29,6 +26,16 @@
...
 
+try:
+    # The bitshuffle plugin needs to be imported at least once in the session
+    # so that h5py can use the plugin.
+    import bitshuffle.h5  # noqa: F401
+except ImportError:
+    pytest.skip(
+        allow_module_level=True,
+        reason="File output regression test cannot be done without `bitshuffle.h5`.",
+    )
```
SimonHeybrock: Why not `pytest.importorskip`?

SimonHeybrock (approved):

packages/essnmx/tests/executable_output_test.py
YooSunYoung: Yeah you're right...! I updated it.

PR #686, opened 2026-07-31
Drop Python 3.11, declare and test up to Python 3.14

MridulS:
- Require Python >= 3.12 in all packages (pixi pin, requires-python, classifiers, copier answers, workflow pins) and re-lock
- Declare Python 3.14 support in classifiers and copier max_python
- Bump minimum scipp to 26.7.0
- Nightly: run latest-dependencies on 3.12/3.13/3.14 and add a free-threaded 3.14t job; essnmx is excluded (bitshuffle re-enables the GIL) as is essdiffraction (numba aborts in dask worker threads)
- Unify uv at 0.12.0 across nightly jobs


packages/essdiffraction/.copier-answers.yml
```diff
@@ -2,8 +2,8 @@
 _commit: 024a41b
```
nvaytet, line 2: Maybe I missed something but we do we still have copier answers? I thought we were not using copier in the monorepo?

packages/essimaging/docs/developer/getting-started.md
```diff
@@ -40,7 +40,7 @@ Alternatively, if you want a different workflow, take a look at ``tox.ini`` or `
 Run the tests using
 
 ```sh
-tox -e py311
+tox -e py312
```
nvaytet, line 43: This documentation seems out of date? We no longer use `tox`. Left over from copier?

packages/essreduce/pyproject.toml
```diff
@@ -17,23 +17,23 @@ classifiers = [
...
+requires-python = ">=3.12"
 
 dynamic = ["version"]
 
 dependencies = [
     "dask>=2022.1.0",
     "graphviz>=0.20",
     "sciline>=25.11.0",
-    "scipp>=26.3.1",
+    "scipp>=26.7.0",
```
nvaytet: What was the reason for the scipp version bump?

packages/essdiffraction/.copier-answers.yml
MridulS: We still have copier.yml lying around, I am happy to remove copier.yml files here if we have fully decided on not using copier in the monorepo. I am not sure if we had a final decision here? cc @YooSunYoung

packages/essreduce/pyproject.toml
MridulS: I don't actually remember, I think there was some local failure with py314 scipp build but I can't reproduce the failure anymore. I'll revert.

packages/essdiffraction/.copier-answers.yml
YooSunYoung: We kept it at the beginning in case we want to use copier template to, e.g. update python version using copier template,,, but if I'm the only one who's willing to maintain the copier template we should just remove them... I don't want to be too important.

In Short: Let's remove them...?

jl-wynen: Yes, but in a separate PR, please.

.github/workflows/nightly.yml
```diff
@@ -143,23 +143,54 @@ jobs:
...
         working-directory: packages/${{ matrix.package }}
         run: uv run --extra=test --resolution=highest pytest
 
+  free-threaded:
+    name: Free-threaded ${{ matrix.package }}
+    strategy:
+      fail-fast: false
+      matrix:
+        # Excluded packages that do not work on free-threaded Python yet:
+        # - essnmx: importing bitshuffle re-enables the GIL, which
```
jl-wynen: Can you suppress that warning specifically in the project? I think that is less likely to cause issues down the line than silently not testing the package.

.github/workflows/nightly.yml
```diff
@@ -143,23 +143,54 @@ jobs:
...
+        #   filterwarnings = error escalates to a collection error.
+        # - essdiffraction: numba-jitted code aborts the interpreter when run
```
jl-wynen: Will #707 fix this?

packages/essdiffraction/.copier-answers.yml
nvaytet: @YooSunYoung I'm not saying the copier template is not important. It's still used in plenty of other repositories. I just thought that they no longer serve a purpose here because I don't think it can work in subdirectories of a repo (e.g. for the docs config), and all the CI setup and dependencies is now managed centrally at the root of the monorepo.

.github/workflows/nightly.yml
MridulS: actually yes, it should

MridulS:
Anymore blockers here?

jl-wynen (commented):
LGTM. Do we need to wait for #707 or can we merge this without degradations?

MridulS:
I think we can merge this for now, the #707 test is anyway a flaky failure.

jl-wynen (approved):

PR #689, opened 2026-08-05
[ESSREDUCE] refactor: WIP

jokasimr:
Copied from #654

### DIFF
--- a/packages/essreduce/src/ess/reduce/parameter.py
+++ b/packages/essreduce/src/ess/reduce/parameter.py
@@ -2,12 +2,11 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 from __future__ import annotations
 
-from collections.abc import MutableMapping
-from dataclasses import dataclass
-from enum import Enum
-from typing import Generic, Self, TypeVar
+from collections.abc import Callable, MutableMapping
+from dataclasses import dataclass, replace
+from typing import Any, Generic, TypeVar
 
-import scipp as sc
+from sciline import Pipeline
 from sciline._utils import key_name
 from sciline.typing import Key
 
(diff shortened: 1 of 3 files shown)


SimonHeybrock (commented):
With a new look, I think a remember our discussion from before summer (on the original PR), as well as some new comments:

1. Concerns about having `model` in `ParameterSpec`. My idea was that `WorkflowSpec` should be independent of workflow *implementation*. Having `model` kind of prescribes a `sciline` implementation.
2. `use_workflow_default` is similar (linked to what we have in our sciline-based workflows). I think our conclusion was that we should never use workflow defaults.
3. I think `title` and `description` should be mandatory.
4. Is `ParameterSpec` necessary, or would it be cleaner to just use `pydantic`? We may alternatively consider using a `Protocol` that could be fulfilled by Pydantic models.
5. `WorkflowSpec`: `typical_outputs` and other rather bespoke fields should likely be kept out of a generic spec, at least in v0. This feature was what we used in the original UI generation code, but should maybe be kept in a subclass for now (not part of this PR).
6. Is `WorkflowRegistry` necessary?
7. Why does `WorkflowSpec` hold a workflow factory? The intent of `WorkflowSpec` is to be generic metadata for a workflow, this seems to much tied to what is implemented concretely (based on old implementation)?

I think it is easier to have a fresh suggestion instead of editing all this into this one. I will open a PR for discussion.

SimonHeybrock:
As announced above, I have opened #690 as a fresh take — pure-interface spec (no factory, no sciline keys), one pydantic params model per workflow, plus an ADR with the rationale. Thanks for the groundwork here; much of the analysis carried over. I propose we close this PR and continue the discussion on #690.

PR #694, opened 2026-08-06
[ESSSANS] docs: compute reference curve instead of reading it from file

### LINKED ISSUE #679: [ESSSANS] Add code for computing PolyGauss instead of using a file
Currently we have `PolyGauss_I0-50_Rg-60.h5` stored in our data. According to Judith this is simple to implement in code (avoiding the need to track and download a file), e.g., using SciPy. Judith will provide code for this.

jokasimr:
Fixes #679 

I saw that Judith would provide code for this, but I thought it was a bit fun to let codex have a stab at it.

Max absolute difference between curve in file and computed curve is about 1e-8, with max relative difference being about 2e-10.


packages/esssans/src/ess/loki/data.py
```diff
@@ -164,6 +166,31 @@ def loki_tutorial_poly_gauss_I0() -> Path:
     return Path(_registry.get_path('PolyGauss_I0-50_Rg-60.h5'))
 
 
+def loki_tutorial_poly_gauss_I0_data() -> sc.DataArray:
+    """Return the analytical I(Q) model for the tutorial Poly-Gauss sample.
+
+    The model is the SasView ``poly_gauss_coil`` model with an intensity at zero
+    Q of 50, a radius of gyration of 60 angstrom, a polydispersity of 1.02, and
+    a background of 0.2.
```
nvaytet, line 174: I think the function would be more useful if all the parameters would be values we can tune?
You can still give them the default values you list here, but it would make it useful in case we want a new intensity at Q=0, different Q range, etc..

SimonHeybrock: Yes, the goal of the issue was to make the computation part of a production code (not just a tutorial helper function), i.e., a graph node computing the model. I don't know if the params should all be workflow params, check with Judidth.

PR #697, opened 2026-08-10
[ESSSPECTROSCOPY] test: add absolute tolerances to make tests less flaky

### LINKED ISSUE #696: Weekly Platform Test Failures 2026-08-10
See https://github.com/scipp/ess/actions/runs/31354093981 for more details.

jokasimr:
Fixes https://github.com/scipp/ess/issues/696

(https://github.com/scipp/ess/actions/runs/31354093981/job/93350340119)

### DIFF
--- a/packages/essspectroscopy/tests/bifrost/workflow_test.py
+++ b/packages/essspectroscopy/tests/bifrost/workflow_test.py
@@ -117,7 +117,12 @@ def test_simulation_workflow_produces_the_same_data_as_before(
         sc.testing.assert_allclose(
             energy_data.coords[name],
             expected.coords[name],
-            atol=sc.scalar(1e-15, unit=expected.coords[name].unit),
+            atol=(
+                sc.scalar(1e-15, unit=expected.coords[name].unit)
+                # TODO: When scipp supports unit=None in .to(), remove this case.
+                if expected.coords[name].unit is not None
+                else None
+            ),
         )
 
     assert energy_data.bins.coords.keys() == expected.bins.coords.keys()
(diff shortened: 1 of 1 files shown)


jokasimr:
The workaround for `unit=None` will not be required after https://github.com/scipp/scipp/pull/3945 has been deployed.

packages/essspectroscopy/tests/bifrost/workflow_test.py
```diff
@@ -114,12 +114,28 @@ def test_simulation_workflow_produces_the_same_data_as_before(
 
     assert set(expected.coords.keys()).issubset(set(energy_data.coords.keys()))
     for name in expected.coords.keys():
-        sc.testing.assert_allclose(energy_data.coords[name], expected.coords[name])
+        sc.testing.assert_allclose(
+            energy_data.coords[name],
+            expected.coords[name],
+            atol=(
+                sc.scalar(1e-15, unit=expected.coords[name].unit)
+                # TODO: When scipp supports unit=None in .to(), remove this case.
```
nvaytet, line 122: So does this fail when one of the coords has `unit=None` because internally `assert_allclose` is calling `.to(unit=)` on the arguments?

Maybe the call to `.to()` should be removed and we should just check if the units are the same? (thus requiring manual conversion from the user before calling `assert_allclose`, but maybe that's ok?)

jokasimr: I think the fix is rather to make `.to()` accept `unit=None`, I have made a PR to Scipp to add support for that.

nvaytet: my question was more: what was the reason for needing to treat the `unit is None` case?

nvaytet: After in-person discussion: calling `.to(unit=...)` is only called on the `atol`, not on the variables passed.
So ok for the workaround here.

nvaytet (approved):

PR #698, opened 2026-08-11
[ESSSPECTROSCOPY] Source the BIFROST Bragg peak monitor from the NXmonitor

SimonHeybrock:
`get_calibrated_bragg_peak_detector` required `NeXusComponent[NXdetector]`, but the Bragg peak monitor is the elastic monitor (cbm5), written as an `NXmonitor` in CODA files, the geometry artifacts and the McStas simulation file alike. Loading therefore raised

```
ValueError: The NeXus group 'elastic_monitor' was expected to be a NXdetector but is a NXmonitor
```

against every real file. Only the user guide worked, by standing a detector triplet in for the monitor.

The monitor's component, transformation and event data now come from `ElasticMonitor`, which the workflow already declares in `monitor_types` — a re-pointing rather than new machinery. A monitor has no pixel offsets, so the position is the transformed origin as in `get_calibrated_monitor`. `Analyzer` is dropped: `get_base_calibrated_detector_bifrost` never reads it, and a monitor in the direct beam has none. The detector stand-in moves to `simulation_providers`, keeping the user guide working.

Assembly likewise uses `assemble_monitor_data`. `assemble_detector_data` groups events by `event_id` onto a `detector_number` grid, and a single-pixel monitor has nothing to group by; what remains is the geometry assignment `assemble_monitor_data` already does. This holds whether or not the events carry an `event_id` — a file-loaded `cbm5_events` group does, constant `1` throughout; a stream may not.

Also fixes a latent defect that is not monitor-specific: `Ltotal` was derived from `EmptyDetector`, a branch of the graph that never sees the grouping, so a time-dependent position left it labelled `time` while `group_by_rotation` had already turned that dimension into `a4`, and the broadcast in `detector_wavelength_data` rejected it. This bites whenever the tank rotates. It stayed hidden because the user guide's stand-in position is static and the inelastic/Q-cut path uses a `detector_wavelength_data` that takes no `ltotal`. `Ltotal` is now computed from the grouped data, so it carries the right dimensions by construction; the user guide's local override, which existed for the same reason, is gone with it.

## Testing

No regression test for the monitor path here — the simulation file's `elastic_monitor` is histogram-mode, so there is no simulated Bragg peak monitor to drive. It is covered end to end in scipp/esslivedata#1236, where streamed monitor events and live rotation readbacks produce a populated (Qpar, Qperp) map over a multi-angle scan. That branch needs this PR, so it is blocked on it.

## Open question

The scientists asked for the map in the laboratory frame, but `project_momentum_transfer` projects `sample_table_momentum_transfer` — the sample-table frame, which is what you want against an expected reciprocal lattice. Worth settling before merge.

### DIFF
--- a/packages/essspectroscopy/src/ess/bifrost/single_crystal/detector.py
+++ b/packages/essspectroscopy/src/ess/bifrost/single_crystal/detector.py
@@ -1,19 +1,90 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2026 Scipp contributors (https://github.com/scipp)
 
-"""Bragg peak detector handling for BIFROST."""
+"""Bragg peak monitor handling for BIFROST."""
 
+import scipp as sc
 import scippnexus as snx
 from ess.spectroscopy.types import (
     Analyzer,
     DetectorPositionOffset,
+    ElasticMonitor,
     EmptyDetector,
     NeXusComponent,
+    NeXusData,
     NeXusTransformation,
+    RawDetector,
     RunType,
 )
 
-from ..detector import get_base_calibrated_detector_bifrost
+from ess.reduce.nexus.types import MonitorPositionOffset
+
+from ..detector import _assign_detector_position, get_base_calibrated_detector_bifrost
+
+
+def get_calibrated_bragg_peak_monitor(
+    monitor: NeXusComponent[ElasticMonitor, RunType],
+    *,
+    transform: NeXusTransformation[ElasticMonitor, RunType],
+    offset: MonitorPositionOffset[RunType, ElasticMonitor],
+) -> EmptyDetector[RunType]:
+    """Extract the data array corresponding to the Bragg peak monitor's signal field.
+
+    BIFROST's Bragg peak monitor is the elastic monitor (``cbm5``), written as an
+    ``NXmonitor``. It has no pixel offsets, so its position is the transformed origin
+    as in :func:`ess.reduce.nexus.workflow.get_calibrated_monitor`, rather than the
+    per-pixel computation used for detectors. The position is assigned with the
+    BIFROST-specific broadcasting because the monitor is mounted on the detector tank
+    and therefore moves with the instrument angle.
+
+    Parameters
+    ----------
+    monitor:
+        Loaded NeXus monitor.
+    transform:
+        Transformation that determines the monitor position.
+    offset:
+        Offset to add to the monitor position.
+
+    Returns
+    -------
+    :
+        Monitor with geometry coordinates.
+    """
+    from ess.reduce.nexus import extract_signal_data_array
+
+    da = extract_signal_data_array(monitor)
+    unit = transform.value.unit
+    position = transform.value * sc.vector([0.0, 0.0, 0.0], unit=unit) + offset.to(
+        unit=unit
+    )
+    # A monitor carries no detector_number; the Bragg peak monitor is a single pixel,
+    # so name it explicitly for the pixel-grouping done by the event assembly below.
+    da = da.assign_coords(detector_number=sc.index(1))
+    return EmptyDetector[RunType](_assign_detector_position(da, position))
+
+
+def assemble_bragg_peak_monitor_data(
+    monitor: EmptyDetector[RunType],
+    data: NeXusData[ElasticMonitor, RunType],
+) -> RawDetector[RunType]:
+    """Combine the Bragg peak monitor's geometry with its event data.
+
+    Parameters
+    ----------
+    monitor:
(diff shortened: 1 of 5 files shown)


packages/essspectroscopy/src/ess/bifrost/single_crystal/time_of_flight.py
```diff
@@ -32,6 +32,13 @@ def detector_wavelength_data(
     :func:`ess.reduce.unwrap.detector_wavelength_data`
     for different input types.
     """
+    # A time-dependent detector position (BIFROST's tank rotates, and the Bragg peak
+    # monitor is mounted on it) makes ``ltotal`` depend on 'time'. The instrument angle
+    # is the only dynamic parameter, so ``group_by_rotation`` has already turned that
+    # same 'time' dimension into 'a4'. Rename to match, or the broadcast below rejects
+    # ``ltotal`` as having a dimension the data does not.
+    if 'time' in ltotal.dims and 'time' not in sample_data.dims:
+        ltotal = ltotal.rename_dims(time='a4')
```
jl-wynen: Is `ltotal` actually grouped by a4 here?

Can we instead compute `ltotal` from the detector after grouping? That way, we would always get the correct dimension.

packages/essspectroscopy/src/ess/bifrost/single_crystal/workflow.py
```diff
@@ -45,6 +46,12 @@ def BifrostBraggPeakMonitorWorkflow() -> sciline.Pipeline:
     )
     # Use the vanilla implementation instead of the indirect geometry one:
     workflow.insert(reduce_unwrap.to_wavelength.detector_wavelength_data)
+    # The Bragg peak monitor sees the direct beam, so its flight path is a straight
```
jl-wynen: What does 'direct beam' mean here? It is not the transmitted beam.

packages/essspectroscopy/src/ess/bifrost/single_crystal/workflow.py
```diff
@@ -45,6 +46,12 @@ def BifrostBraggPeakMonitorWorkflow() -> sciline.Pipeline:
...
+    # line rather than the analyzer-folded path of the inelastic detectors.
+    workflow.insert(
+        reduce_unwrap.to_wavelength.detector_ltotal_from_straight_line_approximation
+    )
```
jl-wynen: With this replacement, there is a risk that it gets used for the detectors. Does this workflow still have providers for detectors or only the elastic monitor?

packages/essspectroscopy/src/ess/bifrost/single_crystal/detector.py
```diff
@@ -1,29 +1,99 @@
...
+
+
+def get_calibrated_bragg_peak_monitor(
+    monitor: NeXusComponent[ElasticMonitor, RunType],
+    *,
+    transform: NeXusTransformation[ElasticMonitor, RunType],
+) -> EmptyDetector[RunType]:
+    """Extract the data array corresponding to the Bragg peak monitor's signal field.
+
+    BIFROST's Bragg peak monitor is the elastic monitor (``cbm5``), written as an
```
jl-wynen: 'cmb5 doesn't really have a meaning in ESSspectroscopy. Can you mention that it is called 'elastic_monitor' in files?

packages/essspectroscopy/src/ess/bifrost/single_crystal/workflow.py
SimonHeybrock: Meant direct from sample, not going through the analyzer.

jl-wynen: Can you adjust the comment so that this doesn't get confused with the definition in SANS?

packages/essspectroscopy/src/ess/bifrost/single_crystal/time_of_flight.py
SimonHeybrock: Good catch, and yes -- `ltotal` came off `EmptyDetector`, a branch of the graph that never sees the grouping, so it could only ever carry the pre-grouping dimension. It now comes from `DataGroupedByRotation`, which gets the dimension right by construction. It also removes an override the user guide needed for the same reason.

packages/essspectroscopy/src/ess/bifrost/single_crystal/detector.py
SimonHeybrock: Fair -- reworded to lead with `elastic_monitor` and keep `cbm5` only as the instrument-documentation name.

jl-wynen (approved):

PR #699, opened 2026-08-11
refactor: move ZoomTransmissionFraction workflows to esssans

### LINKED ISSUE #615: [ESSreduce] The polarization package imports from ESSsans
`src/ess/reduce/polarization/zoom.py` imports a couple of things from modules in ESSsans:
```python
from ess.isissans.io import LoadedFileContents
from ess.sans.types import (
    Filename,
    Incident,
    MonitorType,
    NeXusComponent,
    NeXusMonitorName,
    RunType,
    SampleRun,
    Transmission,
    TransmissionRun,
)
```
This is backwards because polarization is now in ESSreduce. Also, the tests are not failing, so it seems like we are not testing this module.

jokasimr:
Fixes #615


packages/esssans/src/ess/isissans/zoom.py
```diff
@@ -53,3 +73,95 @@ def ZoomTutorialWorkflow() -> sciline.Pipeline:
     workflow.insert(load_tutorial_run)
     workflow.insert(load_tutorial_direct_beam)
     return workflow
+
+
+def _get_time(dg: sc.DataGroup) -> sc.Variable:
```
nvaytet, line 78: Is it correct that the code here was simply copied from the deleted file?

nvaytet (approved):

packages/esssans/src/ess/isissans/zoom.py
jokasimr: Yes that's correct

Issue #669, opened 2026-07-23
[ESSDIFFRACTION, BEER] Add correction to loaded McStas NeXus files


celinedurniak:
- Modify loader of simulated BEER data (NeXus file from McStas) for the pulse shaping modes of operation of the choppers (modes 4-6 in Mcstas convention). See the enclosed notebook for an example of implementation
- Add L2 (sample - detector distance for each pixel in the loaded data array)
- Add option to load all the detector banks at once

[pr_beer_loader_sciline.ipynb](https://github.com/user-attachments/files/30297387/pr_beer_loader_sciline.ipynb)

Input files to run the notebook: duplex sample and Vanadium NeXus files from McStas simulations (mode 4 for the choppers, 500x500 resolution for each of the detector panels) stored on NextCloud: https://project.esss.dk/nextcloud/index.php/f/21246100

PR #704, opened 2026-08-14
[ESSDIFFRACTION] BEER improvements

jokasimr:
Main changes are:

* Adds the modulation chopper modes to the list of chopper settings.
* Uses chopper parameters instead of mcstas specific parameters in modulation workflow, making it more ready for real data. 
* Simplifies the McStas data loader using McstasTox - it needed some refactoring.

PR #706, opened 2026-08-17
[ESSDIFFRACTION] Both banks

jokasimr:
Fixes part of #669 by adding the option to load both banks at once.


jokasimr:
Review should probably wait until  https://github.com/scipp/ess/pull/704 has been merged, because then the diff will be smaller.

jl-wynen (approved):

PR #707, opened 2026-08-17
Fix threading

jokasimr:
Fixes the failing weekly tests on MacOS: https://github.com/scipp/ess/issues/705

### DIFF
--- a/packages/essreduce/src/ess/reduce/unwrap/to_wavelength.py
+++ b/packages/essreduce/src/ess/reduce/unwrap/to_wavelength.py
@@ -7,6 +7,7 @@ This workflow is used to convert raw detector data with event_time_zero and
 event_time_offset coordinates to data with a time-of-flight coordinate.
 """
 
+import warnings
 from collections.abc import Callable
 from dataclasses import asdict
 
@@ -15,11 +16,6 @@ import scipp as sc
 import scippneutron as scn
 import scippnexus as snx
 from scippneutron._utils import elem_unit
-
-try:
-    from .interpolator_numba import Interpolator as InterpolatorImpl
-except ImportError:
-    from .interpolator_scipy import Interpolator as InterpolatorImpl
 
 from ..nexus.types import (
     Component,
@@ -37,6 +33,7 @@ from .resample import rebin_strictly_increasing
 from .types import (
     DetectorLtotal,
     ErrorLimitedLookupTable,
+    FrameUnwrapBackend,
     KeepEventTimeOffset,
     LookupTable,
     LookupTableRelativeErrorThreshold,
@@ -47,6 +44,39 @@ from .types import (
 )
 
 
+def _get_interpolator_class(backend: FrameUnwrapBackend) -> type:
+    if backend == FrameUnwrapBackend.scipy:
+        from .interpolator_scipy import Interpolator
+
+        return Interpolator
+
+    try:
+        from numba import get_num_threads, threading_layer
+
+        # Initialize the threading layer now so we can inspect the selected backend
+        # and fall back before interpolation.
+        get_num_threads()
+        if threading_layer() in {'omp', 'tbb'}:
+            from .interpolator_numba import Interpolator
+
+            return Interpolator
+    except (ImportError, ValueError):
+        pass
+
+    warnings.warn(
+        "The 'numba' frame-unwrapping backend was requested, but Numba is "
+        "unavailable or did not select a thread-safe threading layer. Falling "
+        "back to the 'scipy' backend. This fallback is deprecated and will be "
+        "an error in a future release. Select the SciPy backend explicitly with "
+        "backend='scipy'.",
+        FutureWarning,
+        stacklevel=3,
+    )
+    from .interpolator_scipy import Interpolator
+
+    return Interpolator
+
+
 class WavelengthInterpolator:
     def __init__(
         self,
@@ -54,6 +84,7 @@ class WavelengthInterpolator:
         distance_unit: str,
         time_unit: str,
         wavelength_unit: str = 'angstrom',
+        backend: FrameUnwrapBackend = FrameUnwrapBackend.numba,
     ):
         """
         Interpolator object that converts event_time_offset and distances to
@@ -96,7 +127,7 @@ class WavelengthInterpolator:
 
         distances = lookup.coords["distance"].to(unit=distance_unit, copy=False)
 
-        self._interpolator = InterpolatorImpl(
+        self._interpolator = _get_interpolator_class(backend)(
             time_edges=time_coord,
             distance_edges=distances.values,
             values=(
@@ -137,7 +168,10 @@ class WavelengthInterpolator:
 
 
 def _compute_wavelength_histogram(
-    da: sc.DataArray, lookup: LookupTable, ltotal: sc.Variable
+    da: sc.DataArray,
+    lookup: LookupTable,
+    ltotal: sc.Variable,
+    backend: FrameUnwrapBackend,
 ) -> sc.DataArray:
     # In NeXus, 'time_of_flight' is the canonical name in NXmonitor, but in some files,
     # it may be called 'tof' or 'frame_time'.
@@ -164,7 +198,10 @@ def _compute_wavelength_histogram(
 
     # Create linear interpolator
     interp = WavelengthInterpolator(
-        lookup.array, distance_unit=ltotal.unit, time_unit=eto_unit
+        lookup.array,
+        distance_unit=ltotal.unit,
+        time_unit=eto_unit,
+        backend=backend,
     )
 
     # Compute wavelengths of the bin edges using the interpolator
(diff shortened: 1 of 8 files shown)


SimonHeybrock:
Can you clarify what the implications of this are (performance?), and whether we fallback in practice (production) &mdash; and if so, how we can avoid that. It seems like this might silently introduce performance regressions, depending on environment.

jokasimr:
> Can you clarify what the implications of this are (performance?), and whether we fallback in practice (production) — and if so, how we can avoid that. It seems like this might silently introduce performance regressions, depending on environment.

The implications are that in environments that don't have `tbb` or `openmp` we will not use the Numba interpolation implementation.

This will lead to a regression in environments without those libraries, because those will fall back to the slower `scipy` implementation. But that is almost certainly better than randomly crashing, which is what might happen otherwise.

The required libraries should be available on all our production platforms. It might of course fall back in some user environment that we don't control.

Alternatives:

1. Don't use dask threading. This will make this issue less likely to happen in our typical setups, but we loose dask thread parallelism and the workflows will still be thread-unsafe and anyone running the workflows in a thread pool might experience the same issue.

2. Instead of silent fallback, raise or warn and encourage the user to install a thread safe backend.

SimonHeybrock:
Have you looked into our actual envs &mdash; when you say "should be available", does that mean **can be** installed or **are installed**? What environments do not have `tbb` or `openmp`? Disabling threaded dask for tests might be a less-invasive fix for CI, if we need more time to figure that out.

jokasimr:
>  when you say "should be available", does that mean can be installed or are installed? 

I mean "can be installed".

jokasimr:
> Disabling threaded dask for tests might be a less-invasive fix for CI

Maybe, but then we don't catch issues like this.

jokasimr:
I don't really think this is a big change. The code already falls back to the non-numba implementation if `numba` is not installed, and `numba` is not an explicit requirement of `essreduce`. This PR changes that condition to: "fall back to the non-numba implementation if `numba` with a threadsafe backend is not installed".

SimonHeybrock:
> I don't really think this is a big change. The code already falls back to the non-numba implementation if `numba` is not installed, and `numba` is not an explicit requirement of `essreduce`.

Hmm, does seems odd. Isn't it what the wavelength-LUT code wants to rely on @nvaytet? Should it be added, for `essreduce`, or the technique packages? Do our VISA images have it?

> This PR changes that condition to: "fall back to the non-numba implementation if `numba` with a threadsafe backend is not installed".

Knowing "I have to install numba to get the fast implementation" is relatively easy, knowing that there might be a silent fallback to a slow implementation if some threading library (the user may have never heard of) is not available is arguably worse.

jokasimr:
> Knowing "I have to install numba to get the fast implementation" is relatively easy, knowing that there might be a silent fallback to a slow implementation if some threading library (the user may have never heard of) is not available is arguably worse.

Right now the user still has to remember to install some threading library, or they might see random crashes.

SimonHeybrock:
> > Knowing "I have to install numba to get the fast implementation" is relatively easy, knowing that there might be a silent fallback to a slow implementation if some threading library (the user may have never heard of) is not available is arguably worse.
> 
> Right now the user still has to remember to install some threading library, or they might see random crashes.

In other words, it is not *silent*, that is my point. I think not silently degrading performance is better.

jokasimr:
I've now added a workflow parameter for selecting the wavelength unwrapping backend.

If the value of that parameter is `numba` (the default) then it will check if numba is installed with a thread safe backend, if it isn't then it falls back to using scipy (preserving current behavior), but it will also raise a deprecation warning explaining the issue. That lets us change this to be a hard error in the future.

SimonHeybrock (commented):
Looks like a reasonable approach, what do you think @nvaytet or @jl-wynen ?

packages/essreduce/src/ess/reduce/unwrap/to_wavelength.py
```diff
@@ -47,13 +44,47 @@
 )
 
 
+def _get_interpolator_class(backend: FrameUnwrapBackend) -> type:
+    backend = FrameUnwrapBackend(backend)
```
jl-wynen:
```suggestion
```

It's a noop if the function is called properly. (i.e. passes mypy)

packages/essreduce/src/ess/reduce/unwrap/to_wavelength.py
```diff
@@ -47,13 +44,47 @@
...
+    if backend == FrameUnwrapBackend.scipy:
+        from .interpolator_scipy import Interpolator
+
+        return Interpolator
+
+    try:
+        from numba import config, get_num_threads, threading_layer
+
+        config.THREADING_LAYER = 'threadsafe'
```
jl-wynen: I don't like that this sets a global variable. It will be hard to debug threading behaviour in code that uses the workflow. 
Is this strictly required here?

packages/essreduce/tests/unwrap/interpolator_test.py
```diff
@@ -100,3 +107,65 @@ def test_numba_and_scipy_interpolators_yield_same_results_with_values_on_edges()
...
+def test_numba_interpolator_can_be_called_from_multiple_threads():
+    numba_interp, scipy_interp = _make_interpolators()
+    npoints = 100_000
+    times = np.linspace(0.0, 71.0, npoints)
+    distances = np.linspace(40.0, 70.0, npoints)
+    barrier = threading.Barrier(2)
+
+    def interpolate_after_barrier(_):
+        barrier.wait()
+        return numba_interp(times, distances)
```
jl-wynen: I think you need a second barrier after the call to `numba_interp`. Otherwise, one thread can call the interpolator and finish before the other calls the interpolator.

But in general, there is no guarantee that both threads will be in the interpolator at the same time.

pixi.toml
```diff
@@ -8,6 +8,9 @@ exclude-newer = "14d"
 [dependencies]
 python = "3.11.*"
 
+[target.osx-arm64.dependencies]
+llvm-openmp = "*"
```
jl-wynen: Why is this required? And how does this map onto the wheel and conda package dependencies?

packages/essreduce/tests/unwrap/interpolator_test.py
jokasimr: I don't think the second barrier will help with that because, like you note, the threads will not be in the interpolator at the same time, and I think that is required for the crash to occur.

There's no guarantee the threads will be in the interpolator at the same time, but I think it is quite likely, we could increase the number of points to make it more likely.

We could also just remove the test.

packages/essreduce/src/ess/reduce/unwrap/to_wavelength.py
jokasimr: Instead of modifying `THREADING_LAYER` we could just check that it is one of the threadsafe ones.
I think numba will by default select a threadsafe layer if there is one available.

pixi.toml
jokasimr: It is required because the osx-arm64 platform does not have `tbb` or `openmp` available by default, and then numba will use a non-thread safe backend, and the tests in essreduce will fall back to using the scipy interpolator.

SimonHeybrock: Given that most of ESS is using Macs, should we add this to our package requirements?

jl-wynen: But this only affects dev environments. The package deps are in `pyproject.toml`.

jokasimr: `pyproject.toml` cannot install a conda package, so as I understand it we can not make this part of the dependencies of `essreduce`?

However, we can make sure to add this to the ESS environment that is deployed on VISA and that users can download to reproduce the environment locally.

SimonHeybrock: @MridulS What do you think? We'd also need to make sure this is installed in the ESSlivedata backends for example, which just use `pip`.

MridulS: This shouldn't have any affect on visa or livedata as they are anyway running on linux boxes, but I'm not a 100% sure if this is going in the right direction. Something seems off, where is windows/linux getting tbb/openmp from then? Discussing this with Johannes.

MridulS:
https://github.com/scipp/ess/blob/main/pixi.lock does have _openmp_mutex for win64, and linux-64. Not for macos ....

MridulS:
After some digging found this https://github.com/numba/numba/issues/10492, looks like putting `numba` as a dependency of essreduce test extras (which forces to pull numba in from pypi) messes this up. The conda-forge numba correctly installs the threading backends.

pixi.toml
MridulS: I think the cleaner solution here for now is to add this to our pixi.toml

```
[feature.essreduce.dependencies]
numba = ">=0.63"  # Prefer conda-forge Numba as it installs thread safe backends for all platforms
```

pixi solves the conda packages first, so this numba should take precedence over the one defined by the `test` optional deps in essreduce pyproject.toml

jokasimr:
Is there anything more to do here? @SimonHeybrock Did you have any concerns or some other preferred solution?

SimonHeybrock:
Probably fine, but what about https://github.com/scipp/ess/pull/707#discussion_r3824659650?

jokasimr:
> Probably fine, but what about [#707 (comment)](https://github.com/scipp/ess/pull/707#discussion_r3824659650)?

That suggestion was implemented the same day the comment was made.

SimonHeybrock (approved):

PR #709, opened 2026-08-18
[ESSSPECTROSCOPY] deps: bump essreduce

### LINKED ISSUE #708: Nightly Failures 2026-08-18
See https://github.com/scipp/ess/actions/runs/32092151823 for more details.

jokasimr:
Fixes #708

But requires `pixi lock` before merge, and `pixi lock` requires `essreduce/26.8.0` on conda-forge.

### DIFF
--- a/packages/essspectroscopy/pyproject.toml
+++ b/packages/essspectroscopy/pyproject.toml
@@ -32,7 +32,7 @@ requires-python = ">=3.11"
 dependencies = [
     "dask>=2022.1.0",
     "graphviz>=0.20",
-    "essreduce>=26.6.0",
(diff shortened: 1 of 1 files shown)


MridulS (approved):
26.8.0 is on conda-forge, but pixi lock should be fine? essreduce is a local package, so it should be anyway resolve to the one in the pixi workspace.

jokasimr:
> 26.8.0 is on conda-forge, but pixi lock should be fine? essreduce is a local package, so it should be anyway resolve to the one in the pixi workspace.

Aha I see, so we only need to update `pixi.lock` when changing other (non-`ess`) dependencies?

PR #716, opened 2026-08-20
[CI] ci: add ai workflow for fixing nightly failures

jokasimr:
This is a suggestion for how we could set up a workflow that lets claude investigate nightly test failures and suggest a fix.


SimonHeybrock (commented):
This seems very complicated. Why not just tell Claude to open a PR if it finds a fix, else and issue with results of investigation?

jokasimr:
> This seems very complicated. Why not just tell Claude to open a PR if it finds a fix, else and issue with results of investigation?

We could do that. (I think I would prefer that too.)

The advantage of this approach is that the agent only has read-only access to the github repo, except for the one PR that is opened automatically if it leaves changes. :shrug:

jokasimr:
Okay I think the new version @SimonHeybrock suggested where the surrounding logic (context creation, PR submission, etc) is removed and the agent is told to handle it is much better.

jokasimr:
It would be nice to get this experiment up and running. We will likely have to tweak things based on how it does.

SimonHeybrock (approved):
Seems reasonable. I suspect we will get cases where Anthropic API is down (or similar issues), resulting in silent Nightly fails that receive neither issue nor PR, but we can address that later based on how often it happens.

PR #721, opened 2026-08-21
[ESSDIFFRACTION] docs: update beer user guide notebook thumbnails

jokasimr:
Currently the beer user guide notebook thumbnails don't look so good, and the modulation mode user guide thumbnail is reused for both notebooks.

This PR refactors the thumbnail creation and adds a separate thumbnail for the beer powder diffraction user guide notebook.


packages/essdiffraction/docs/_static/thumbnails/beer_mcstas_dark.svg
jokasimr, line 1: The pattern is changed because we're using the newer silicon files instead of the old duplex files.
That matches the content of the notebook.

jl-wynen (approved):

